PR #172, opened 2017-09-01
Add FreeDNS (freedns.afraid.org) client installation for Raspbian

Pant:
This is an installation script for FreeDNS (http://freedns.afraid.org/) client for Raspbian. The client is basically a Bash script. It detects the difference in the registered IP on the record with the current IP address of the Raspbian board. If there is a difference, it runs a wget command and updates the record with the current IP address.


nachoparker:
great first contribution, looking forward to many more!!

PR #207, opened 2017-09-18
Initial commit of ncp-web-wizard.

Pant:
This is the front end of a web wizard. It is written in HTML/CSS/JS.

Libraries:
1. [twitter-bootstrap-wizard
](https://github.com/VinceG/twitter-bootstrap-sues)
2. Bootstrap
3. Jquery

Note1: I didn't touch the color palette of bootstrap but I think we should make it look like the rest of ncp-web. All we have to do actually is change the color of the tabs.

Note2: Ajax has to be implemented in order to be able to trigger actions on the backend.

Note3: The array `dataTable` is made in a way to be able to save the progress of the user in a config file (backend task), in order for if something doesn't work, the user can remove sensitive parts of the config file and then attach it to the forum, so we can determine what the user has done at the wizard and be able to help better. This is also a kind feature request for the entire ncp-web which could coexist with this wizard.

 Note 4: The steps are made by having in mind the minimum procedures a user has to make to have a working Nextcloud instance. Thus it is a straight forward wizard that doesn't give as many options as ncp-web does. 

 Update 1: I don't know which of the files in `bootstrap` folder are needed, so we have to investigate on that too and remove the unneeded ones.


nachoparker:
Implemented in `v0.31.0`. You can see my modifications here https://github.com/nextcloud/nextcloudpi/commit/a9f90b9ef087ad2d67fe91eecae9eaaa234f632f

Thank you!

Issue #132, opened 2017-08-20
ncp-web: password prompt for authentication


nachoparker:
Maybe using php-pam?

This would enhance security and also prevent from attacks even if the attacker obtains the cookie

testbird:
Ah, thanks for pointing me to the correct issue item.

The top solution from
https://serverfault.com/questions/45278/authenticate-in-apache-via-system-account
with the current syntax from the comments and adjusted to "require user pi", should do it. (Browser asking user for the credentials.)

Do you have a pointer to how such ncp updates are rolled out?

nachoparker:
cool, thanks very much!!

That solution seems like it would do at least for now. I will point the person that offered to help with this to this thread to see if he has also investigated and what his opinion is on this.

Maybe he has already implemented that same thing.

> Do you have a pointer to how such ncp updates are rolled out?

What do you want to know exactly? NCP checks for newer versions, and then updates them by executing [update.sh](https://github.com/nextcloud/nextcloudpi/blob/master/update.sh)

testbird:
Thanks, very god pointers to get an idea. I don't have enough background yet to write the update snippets, though.

Anyway, here is what configured a password box to pop up for me, now:
```
apt-get install libapache2-mod-authnz-external
a2enmod authnz_external

```

modified /etc/apache2/sites-available/ncp.conf:
```
Listen 4443
<VirtualHost _default_:4443>
  DocumentRoot /var/www/ncp-web
  SSLEngine on
  SSLCertificateFile      /etc/ssl/certs/ssl-cert-snakeoil.pem
  SSLCertificateKeyFile /etc/ssl/private/ssl-cert-snakeoil.key

  <IfModule mod_authnz_external.c>
    DefineExternalAuth pwauth pipe /usr/sbin/pwauth
  </IfModule>

</VirtualHost>


<Directory /var/www/ncp-web/>

  AuthType Basic
  AuthName "Login"
  AuthBasicProvider external
  AuthExternal pwauth

  <RequireAll>

   <RequireAny>
      Require host localhost
      Require local
      Require ip 192.168
      Require ip 10
   </RequireAny>

   Require user pi

  </RequireAll>

</Directory>
```
and a
`apachectl restart`


-----

PS: not sure now if any of these modules really had to be enabled manually
```
a2enmod authn_base
a2enmod authn_core
```

testbird:
[...]

Issue #130, opened 2017-08-20
ncp-web: button to enable SSH


nachoparker:
It would be useful, so people do not need to plug in keyboard and HDMI (or do the `/boot/ssh` trick) to enable SSH.

This would be reasonable to do once web authentication is in place, for security reasons

testbird:
Ok, if with ncp-web now only installing additional services requires SSH, I'd think to require physical sdcard manipulation to obtain SSH access is an adequate security measure.

nachoparker:
Implemented in `v0.38.0`

PR #213, opened 2017-09-23
ncp-web authentication prompt

Pant:
This  commit is based on @testbird post on #132 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,19 +6,20 @@
 
 
 nextcloudpi: nextcloud
-	docker build . -f docker/nextcloudpi/Dockerfile       -t ownyourbits/nextcloudpi:latest
+	docker build . -f docker/nextcloudpi/Dockerfile   -t ownyourbits/nextcloudpi:latest
 
 nextcloud: lamp
-	docker build . -f docker/nextcloud/Dockerfile         -t ownyourbits/nextcloud:latest
+	docker build . -f docker/nextcloud/Dockerfile     -t ownyourbits/nextcloud:latest
 
 lamp: miniraspbian
-	docker build . -f docker/lamp/Dockerfile           -t ownyourbits/lamp-arm:latest
+	docker build . -f docker/lamp/Dockerfile          -t ownyourbits/lamp-arm:latest
 
 miniraspbian: 
-	docker build . -f docker/miniraspbian/Dockerfile   -t ownyourbits/miniraspbian:latest
+	docker build . -f docker/miniraspbian/Dockerfile  -t ownyourbits/miniraspbian:latest
 
 devel: 
-	docker build . -f docker/devel/Dockerfile   -t ownyourbits/nextcloudpi:devel
+	docker build . -f docker/devel/Dockerfile         -t ownyourbits/nextcloudpi:devel
+
 # License
 #
 # This script is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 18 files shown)


nachoparker:
Thanks @Pant @testbird,

I am testing this right now and making a couple little changes.

I am starting to think that file based auth could be a bit more secure.

In case we continue with PAM... maybe we should create a new user without login access and zero permissions just to use the password for this.

Thoughts?

nachoparker:
At this moment it is implemented in devel. I had some fixes to do for it to work: enable some extra modules and installing the pwauth package, very small things.

See https://github.com/nextcloud/nextcloudpi/commit/016df8c64f2b580cfb70e20d1578f05c3f639b9d

I will close the PR but we can continue discussing

testbird:
Advantage of PAM is that when user sets a secure password the login password is secured (changed from the default) at the same time.

testbird:
Prevents user forgetting to change the login account as well.

nachoparker:
yes, it's nice.

I had to disable HTTP2 push though, because it was causing the login prompt to repeat 4 times :(

Not a big deal. Now I am wondering if we should provide a way to change the password for _pi_ in ncp-web.

Also I am unsure about the idea of being able to enable SSH from ncp-web now that it is a bit more secure. Maybe force to change the password in that case would be enough

#130

testbird:
You mean automaticly activating ssh when proper password is set?
Well, regular users won't hardly need ssh, and one would need a way to disable it again anyway.  A basic ncp-web switch should do.

nachoparker:
no, I mean this https://github.com/nextcloud/nextcloudpi/issues/130

I mean being able to activate SSH from ncp-web, so you don't need to plug in a TV and keyboard or do the `/boot/ssh` file trick to activate SSH in NCP.

testbird:
Well, then I have not understood yet what exactly you are unsure about. I'd say a web switch to turn SSH on and off would be nice to have, of course. (By default, SSH can continue to stay off after the install. Meanwhile the admin may create the file /boot/ssh on the SD-Card to enable it for an installation.)

nachoparker:
well you know, it makes for a bigger attack vector. If there's a vulnerability in ncp-web that can be exploited it could be used to gaing SSH access and the pi user has access to passwordless sudo by default.

Do you have telegram? If you would like to continue discussing stuff maybe you would like to join the telegram NCP-dev group. It is nicer to talk over there for actual conversations.

testbird:
Ah right, larger impact, understood. Well, you may better know what kind of users may still need ssh. I wasn't sure but if creating a file on the sdcard (physical access) also works after the installation is completed, itl seems ok as a (documented) hurdle to me (not requireing TV-cable/keyboard hardware).

(Not using it, but you may as well try the [deltachat-android](http://github.com/deltachat/deltachat-android) IMAP-push client to ease email chatting. See https://delta.chat/en/features.)

nachoparker:
didn't know deltachat, sounds great. Unfortunately we already are using the telegram channel.

testbird:
Took me a while to understand, too. With an email chat client one just relives the use of the federated email system, by making it much more mobile accessible, regardless of other central server channels one may use. For example, it may substantially reduce the required delta to answer the notification email that github sends out to you with this text.

testbird:
Fortunately, we are already all using the e-mail channel. :-)

nachoparker:
:D

yeah I'll probably have a look at it one day. Anyway, if you are interested you can tell me. We discuss and share opinions about NCP's features, design and stuff.

Issue #28, opened 2017-06-14
Option for using duckdns?


klundry:
I prefer to use duckdns.org for my dynamic dns, as it doesn't require me to log in and confirm that I'm still using it every so often and it is free.  It has very good documentation and now supports TXT record for use with letsencrypt.  [DuckDNS spec](http://www.duckdns.org/spec.jsp). Is it possible to get support for it added along with No-IP?

nachoparker:
Implemented in `v0.30.0`

PR #217, opened 2017-09-25
Create nc-duckdns.sh

stratacast:
Add script to handle DuckDNS

### DIFF
--- a/etc/nextcloudpi-config.d/nc-duckdns.sh
+++ b/etc/nextcloudpi-config.d/nc-duckdns.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+
+# DuckDNS installation on Raspbian for NextcloudPi
+#
+#
+# Copyleft 2017 by Courtney Hicks
+# GPL licensed (see end of file) * Use at your own risk!
+#
+
+ACTIVE_=no
+DOMAIN_=yourduckdnsdomain
+TOKEN_=your-duckdns-token
+INSTALLDIR=duckdns
+INSTALLPATH=/etc/$INSTALLDIR
+CRONFILE=/etc/cron.d/duckdns
+DESCRIPTION="Free Dynamic DNS provider (need account from https://duckdns.org)"
+
+install() { :; }
+
+
+configure() 
+{
+
+   if [[ $ACTIVE_ == "yes" ]]; then
+      mkdir $INSTALLPATH 2> /dev/null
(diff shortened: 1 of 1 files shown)


etc/nextcloudpi-config.d/nc-duckdns.sh
```diff
@@ -0,0 +1,77 @@
...
+
+configure() 
+{
+
+   if [[ $ACTIVE_ == "yes" ]]; then
+      mkdir $INSTALLPATH 2> /dev/null
+      # Creates duck.sh script that checks for updates to DNS records
+      touch $INSTALLPATH/duck.sh
+      touch $INSTALLPATH/duck.log
+      echo -e "echo url=\"https://www.duckdns.org/update?domains=$DOMAIN_&token=$TOKEN_&ip=\" | curl -k -o $INSTALLPATH/duck.log -K -" > $INSTALLPATH/duck.sh
```
nachoparker: the field `ip=` is empty, is that correct?

Seems like this downloads a bash script from duckDNS? correct?

etc/nextcloudpi-config.d/nc-duckdns.sh
```diff
@@ -0,0 +1,77 @@
...
+DESCRIPTION="Free Dynamic DNS provider (need account from https://duckdns.org)"
+
+install() { :; }
+
+
+configure() 
+{
+
+   if [[ $ACTIVE_ == "yes" ]]; then
+      mkdir $INSTALLPATH 2> /dev/null
```
nachoparker: I'll change it to `mkdir -p $INSTALLPATH`, which is normally cleaner

nachoparker:
Those notes are so that you understand the changes that I will make in my next commit. They are very small things, the script looks great.

Thank you very much! happy to have a new code contributor! Specially this one request, because duckDNS is the oldest feature request in Github!! #28

etc/nextcloudpi-config.d/nc-duckdns.sh
stratacast: There is no need to enter a value for "ip=" in the URL provided by DuckDNS, they handle all of that for you

The download I believe is to just grab OK or KO from them, which will tell you that DuckDNS is working as it should or not

stratacast:
Thank goodness I've been using it for over a year to know how to contribute this! :)

PR #218, opened 2017-09-26
Fixed mistake in duckdns script

stratacast:
Made a mistake on creating the cron job, I don't know how to submit single line edits yet so...here's a whole new pull request, sorry :)


nachoparker:
applied, thanks

PR #221, opened 2017-09-26
nc-swapfile systemd service and bug fix

Pant:
Created a systemd service which runs:
```
ExecStart=/sbin/dphys-swapfile setup
ExecStartPost=/sbin/dphys-swapfile swapon
ExecStop=/sbin/dphys-swapfile swapoff
```
This systemd service requires nc-automount.service so it doesn't run without nc-automount.

Also fix a bug when checking if the current swap file's path is the same with the requested from the user. That way if a user rerun the entry nc-swapfile it would run the script and delete the original swap file.


etc/nextcloudpi-config.d/nc-swapfile.sh
```diff
@@ -26,8 +26,27 @@ is_active()
...
+Requires=nc-automount.services
+After=nc-automount.service
+
+EOF
+
+  systemctl daemon-reload
+
   local ORIG=$( grep -oP "CONF_SWAPFILE=.*" /etc/dphys-swapfile | cut -f2 -d= )
-  [[ "$ORIG" == "$SWAPFILE_" ]] && return
+  [[ "$ORIG" -ef "$SWAPFILE_" ]] && return
```
nachoparker: I just tested this three times and it does not fail for me with the original ==, but I still think that new way is better, let's keep it

Issue #270, opened 2017-10-27
Request: DynDNS support for accounts at spdns.de


TimmThaler:
There are already a lot of dyndns services that can be configured within the setup.
Nevertheless, I really would appreciate, if you could add a script for accounts at *spdns.de*

I already found a python script by *gbausch* here on github and perhaps you could integrate that dyndns service, too.

Here come the link to the python script: https://github.com/gbausch/spdnsUpdater

Thank you very much.

nachoparker:
Hi,

Well, I don't know how popular that service is but it is fine by me to include it.

The only thing is that somebody that actually uses that service needs to maintain it, I can't keep working things that I don't use ;)

PRs are welcome, see some [docs on adding NCP options](https://github.com/nextcloud/nextcloudpi/wiki/Developer-guide), and ask anything ;)

TimmThaler:
Hey,
I don't know how popular that service is, either, but for german users it might be very interesting, because this provider follows the german laws and restrictions for privacy etc.

The script I mentioned in the opening post has been updated over a year ago. So actually there were no changes in the last year.
Perhaps it would be possible to regularly check for changes in the scipt's repository and then update or copy from there, as I think, the original creator uses his own script and wants it up to date and working.

When I find some time, I will read the guide for the NCP options 😄

nachoparker:
cool, it's easy to add something to NCP, you can check the other DDNS services as a starting point. 

Ask anything :D

TimmThaler:
Alright, I'll do my best 😆

TimmThaler:
Okay,
I tried to create a shell script from the original python script.
It works on my computer very well.

I also tried to write a script for the setup by taking the [duckDNS.sh](https://github.com/nextcloud/nextcloudpi/blob/master/etc/nextcloudpi-config.d/duckDNS.sh) as a basis for that.

Perhaps somebody could test my setup file and tell if it works. I would appreciate that!

Load my spdnsUpdaterNCP.sh script here:
https://github.com/TimmThaler/spdnsUpdater

nachoparker:
Nice!

Looks good... two little things

The NCP extra needs to be one single file, so you can either download the python script in the install() step or, even better, include it in the install step as a HERE document

Have you run it from nextcloudpi-config or ncp-web?

TimmThaler:
Hey,
okay, I changed that.
Originally I thought, I could download the script every time new when the updater is set up new to always have an up-to-date version.

[...]

PR #275, opened 2017-11-02
added spdns.de DynDNS service

TimmThaler:
please look here: https://github.com/nextcloud/nextcloudpi/issues/270


nachoparker:
thanks, I am testing it now (without an account)

looks great, only thing I added so far is a restart of the cron daemon so that it picks up the changes

Issue #299, opened 2017-11-16
Fail2ban script bug


Iolaum:
### Expected behavior

Nextcloud config starts without errors from shell

### Actual behavior

After we connect through ssh:

```
$ sudo nextcloudpi-config 
/usr/local/etc/nextcloudpi-config.d//fail2ban.sh: line 67: syntax error near unexpected token `}'
/usr/local/etc/nextcloudpi-config.d//fail2ban.sh: line 67: `}'
# Exit the GUI from the shell to see the error messages above.
$
```

### Steps to reproduce, from a freshly installed image

See Actual behaviour.

### Notes: 

1. As a result of this [lines](https://github.com/nextcloud/nextcloudpi/blob/634bc335d05db8ea06999a38fd10133f9101dd94/etc/nextcloudpi-config.d/fail2ban.sh#L62-L66) of fail2ban install code will not execute when a user tries to set it up.

2. I believe the issue was caused by 6b9699c870cf4 and I am sending a PR with a fix.

nachoparker:
Many thanks for reporting and for the PR, actually I accidentally removed a line, so it's fixed now

https://github.com/nextcloud/nextcloudpi/commit/c4a0bba8170c118e4941f98b3d6af26ec8d6a5c9

PR #300, opened 2017-11-16
fail2ban fix

Iolaum:
Fix for #299
The issue was caused by [this update](https://github.com/Iolaum/nextcloudpi/commit/6b9699c870cf48458251095798544230fc7c3537#diff-996ed172e00fb26a6fad57660adc328d) on /etc/nextcloudpi-config.d/fail2ban.sh and results in fail2ban install process not fully completing because a remaining `}` ends `install` function prematurely.

P.S. Targeted master instead of devel branch because the commit with the bug is not present there but the image, from master, is out on the users.

### DIFF
--- a/etc/nextcloudpi-config.d/fail2ban.sh
+++ b/etc/nextcloudpi-config.d/fail2ban.sh
@@ -56,8 +56,8 @@ service fail2ban start
 
 exit 0
 EOF
-    chmod +x /etc/services-available.d/100fail2ban
-  }
+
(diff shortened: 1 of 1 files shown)


nachoparker:
Thanks for the PR. I accidentally removed a line. Fixed now in 

https://github.com/nextcloud/nextcloudpi/commit/c4a0bba8170c118e4941f98b3d6af26ec8d6a5c9

Issue #289, opened 2017-11-13
File not found when updating.


Iolaum:
### PROBLEMS

When updating nextcloudpi through ssh, file `/etc/apache2/sites-enabled/ncp.conf` was missing. Not sure if this is actually a problem (feel free to close if it's not), just putting it here in case it is.

### Expected behavior

Normal update.

### Actual behaviour

Update proceeded with the following logs.

```
NextCloudPi v0.32.3 is outdated
update to v0.34.6 through 'nextcloudpi-config' or type 'sudo ncp-update'
pi@nextcloudpi:~ $ sudo nextcloud-config


Launching nc-update
Downloading updates
Performing updates
grep: /etc/apache2/sites-enabled/ncp.conf: No such file or directory
sed: can't read /etc/apache2/sites-enabled/ncp.conf: No such file or directory
vm.overcommit_memory = 1
installing redis...
Hit:1 http://archive.raspberrypi.org/debian stretch InRelease
Get:2 http://mirrordirector.raspbian.org/raspbian stretch InRelease [15.0 kB]
Fetched 15.0 kB in 1s (9,203 B/s)   
Reading package lists... Done
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'php-redis' instead of 'php7.0-redis'
The following additional packages will be installed:
  php-igbinary redis-tools
Suggested packages:
  ruby-redis
The following NEW packages will be installed:
  php-igbinary php-redis redis-server redis-tools
0 upgraded, 4 newly installed, 0 to remove and 0 not upgraded.
Need to get 840 kB of archives.
After this operation, 2,435 kB of additional disk space will be used.
Get:1 http://raspbian.mirror.uk.sargasso.net/raspbian stretch/main armhf php-igbinary armhf 2.0.1-1 [21.0 kB]
Get:2 http://raspbian.mirror.uk.sargasso.net/raspbian stretch/main armhf php-redis armhf 3.1.1-1 [97.9 kB]
Get:3 http://raspbian.mirror.uk.sargasso.net/raspbian stretch/main armhf redis-tools armhf 3:3.2.6-1 [381 kB]
Get:4 http://raspbian.mirror.uk.sargasso.net/raspbian stretch/main armhf redis-server armhf 3:3.2.6-1 [341 kB]
Fetched 840 kB in 1s (757 kB/s)       
Selecting previously unselected package php-igbinary.
(Reading database ... 41042 files and directories currently installed.)
[...]

PR #308, opened 2017-11-19
Using sites-available instead of sites-enabled

Iolaum:
Update related to #289 

Instead of working on the symlink `/etc/apache2/sites-enabled/ncp.conf` we work directly with `/etc/apache2/sites-available/ncp.conf` This way if a user has disabled ncp-web - which removes the symlink that would resolve the files path - the operation still progresses without problems as intended.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -165,10 +165,10 @@ exit $RET
 EOF
 
   # 2 days to avoid very big backups requests to timeout
-  test -f /etc/apache2/sites-enabled/ncp.conf && {
-    grep -q TimeOut /etc/apache2/sites-enabled/ncp.conf || \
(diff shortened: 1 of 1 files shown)


Iolaum:
Rebased branch for new changes in master and branched off from there instead of devel since the last commit is in progress work (and changed since I craeted the PR). Not sure if you want me to target master instead of devel for this PR.

nachoparker:
Hi,

Thanks for your PR. You are right.. it is better to check the original than the symlink. I will include as it is.

We always target devel, don't worry about that, I'll adapt it ;)

In the future, it would be good if you asked me before working on a PR, just say it because some of the things in github need discussion and may never be implemented, so it's good to talk about it first before starting to work on them, I wouldn't want to have people work for nothing!

Thanks again for the help!

PR #354, opened 2018-01-11
Added option to install Beta

ovpc:
If BETA=yes it adds "pre"  to the download path. ${PREFIX}
Haven't tested this yet, will do so shortly after creating pull request

### DIFF
--- a/etc/nextcloudpi-config.d/nc-nextcloud.sh
+++ b/etc/nextcloudpi-config.d/nc-nextcloud.sh
@@ -123,11 +123,15 @@ GRANT ALL PRIVILEGES ON nextcloud.* TO $DBADMIN@localhost;
 EXIT
 EOF
 
+  ## IF BETA SELECTED ADD "pre" to DOWNLOAD PATH
+if [[ "BETA_" == yes ]]; then
+    local PREFIX="pre"
+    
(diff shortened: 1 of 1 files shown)


ovpc:
the ${PREFIX} some doesnt work, although 

`pi@nextcloudpi:~ $ echo wget -q https://download.nextcloud.com/server/${PREFIX}releases/nextcloud-$VER_.tar.bz2
wget -q https://download.nextcloud.com/server/releases/nextcloud-.tar.bz2`

When manually adding pre to download PATH /prereleases/ the install works!

testbird:
> If BETA=yes it adds "pre"

I see only [[ "BETA_" == yes ]] in the diff (why are there so many ending "_"s used in this script?).  However, here, that is just a string  and I don't see how this test would ever be true. The script would need to inherit and evaluate something like a ${BETAINSTALL} variable from somewhere.

nachoparker:
if a variable ends in "_" then it will be changeable from ncp-web and nextcloudpi-config.

See the [dev docs](https://github.com/nextcloud/nextcloudpi/wiki/Developer-guide)

nachoparker:
applied, thanks... I'll tweak the typo and test it

Issue #290, opened 2017-11-13
multi languages for ncp?


JimmyKater:
hey nacho & crew ;-)

ncp is still growing in features and reach the state of "devine", i guess ;)

and so it's difficult to improve it or coming up with random thoughts about improvement...
what about this one (and as usual i dunno how difficult it might be to implement it)...
- getting ncp translated into different languages using transifex (like nextcloud server and several apps do)?

i know of several users who don't feel well doing it all in english and so running and setting ncp up in their native tongues would be of an enormous help for them.

this is clearly an enhancement. but i only can mark this one as an issue

cheers
jimmy # #

nachoparker:
thanks for he suggestion

@Espina2 also keeps saying that this is important, if this keeps growing it will be something we'll have to do. 

I don't know how much work that would be either, but it seems like a lot :P

JimmyKater:
as i still don't really know how this works i could only guess...
you need to have all "texts" as variables to be stored in a file.
but i'm sure the guys of nc will be helping you out in case of questions... and i'll be helping you out in case of translations to german ;)

nachoparker:
> i'll be helping you out in case of translations to german ;)

thanks! that would be great

nachoparker:
we have spanish, portuguese, greek, english and german so far ;)

Daniel-Krebs:
If you need a german translation you can count on me too ;)

nachoparker:
finally some work is being contributed in this direction #372 

thanks to @theCalcaholic

chiyi4488:
or chinese ,haha

nachoparker:
> or chinese ,if you need.

great! the code for multilanguage is already out. We need to create some guide or mechanism to help people contribute with translations.

At this moment we have only German

nachoparker:
The code is out.

Thereis a small [guide](https://github.com/nextcloud/nextcloudpi/wiki/Add-a-new-language-to-ncp-web) in the wiki.

theCalcaholic:
Btw, I'm not sure if the character encoding is set up correctly for languages with non-latin characters yet... If any characters are displayed wrong for you, please let us know.

nachoparker:
This has been implemented for a while. I was thinking to show the language dropdown on the top bar, with a link/message to attract translators.

That will be more visible than this github issue

cervorange:
Hello,

I'm interesting to help and try to translate in french but i don't know how to do this.
[...]

PR #372, opened 2018-01-25
localization

theCalcaholic:
This PR adds localization support to the ncp panel (issue #290).
localization files are modular - those for the core application reside in _ncp-web/l10n/_. The localization files for the different apps are loaded from _/usr/local/etc/nextcloudpi-config.d/l10n/<app_name>/_.

The localization files should be compatible with transifex (or could be made compatible with little effort).

The language to use will be read from the browser configuration (from `$_SERVER["HTTP_ACCEPT_LANGUAGE"]` to be specific) but can be overwritten in the nc-webui settings. Additionally I added support for dropdown menus in the NC Panel in order to implement these configuration options.
I  already translated some parts of the interface to German, but it's not completed yet.


ncp-web/ncp-launcher.php
```diff
@@ -86,7 +103,11 @@
 
   foreach( $params as $name => $value) 
   {
-    preg_match( '/^[\w.,@_\/-]+$/' , $value , $matches )
+    if( is_array($value))
+    {
+      $value = "[". join(",", $value) ."]";
+    }
+    preg_match( '/^[\[\]\w.*,@_\/-]+$/' , $value , $matches )
```
nachoparker: what is the reason to add these new special characters?

theCalcaholic:
I added an option to nc-webui. Please note that I didn't finish translation to German yet (not all apps are fully translated). But it should suffice as a demo.

theCalcaholic:
I rebased the PR on the latest devel/master

ncp-web/.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
```
nachoparker: can we get rid of this file? you might want to add this to your gitignore

etc/nextcloudpi-config.d/nc-webui.sh
```diff
@@ -14,6 +14,7 @@
 #
 
 ACTIVE_=yes
+LANGUAGE_=[*auto*,en,de]
```
nachoparker: can we find a different character other than `*`? as you well know we have to improve our shell backend, but until then I would rather only allow characters that are not special characters for bash

ncp-web/.idea/codeStyles/codeStyleConfig.xml
theCalcaholic: Ah, of course. Must have missed it.

etc/nextcloudpi-config.d/nc-webui.sh
theCalcaholic: Of, good catch. Any idea what to use? I couldn't name any character and be sure that it has no special meaning in bash (I'm not experienced enough with the language).
Possible candidates (that look fine in the last definition) are for example:
~ < > # % §

nachoparker: let's use underscores `_` they are just like any other character

ncp-web/l10n/de_DE.json
```diff
@@ -0,0 +1,39 @@
+{"translations": {
```
nachoparker: I think It would be better to somehow include the ncp-apps name translations in the `l10n/nc-app` subfolder, to keep everything related to any app together

etc/nextcloudpi-config.d/nc-webui.sh
theCalcaholic: Well okay. I don't like underscores so much (because they can't be used for the value itself anymore), but since it's only a temporary representation anyway, I'll just use them for now. :+1:

nachoparker:
It works great, thank you very much.

One last thing... could you get rid of commit https://github.com/nextcloud/nextcloudpi/pull/372/commits/e589cd194f06b3f0a4ef29dcff08aa9a1e670eb4 ?

In the end I removed it from `devel`

Then, I'll merge it ;)

Issue #365, opened 2018-01-21
No internet connectivity


SukerYan:
Paste the output of `sudo ncp-report`
I very like Nextcloudpi. I don't report a issue here, just a suggestion. I know the scripts will check the internet status by ping google.com. Unfortunately I am in China. We can't visit any google service. Is it possible to change another domain to check the internet? Such as amazon.com, this is ok for Chinese people. Thanks.

tflidd:
You can replace google.com by any other domain:
https://github.com/nextcloud/nextcloudpi/search?q=google.com&type=Code&utf8=%E2%9C%93

We could ping different services and the setup is only considered offline when all fail. Or at least take some service that is not blocked anywhere.

SukerYan:
@tflidd Thanks.

Pant:
I made a pull request #389 changing 'google.com' to 'nextcloud.com'.

nachoparker:
wow didn't think of that... originally I was pinging github, but then it seems like it filters ping in some countries, like Germany, so I changed to google...

another option is to check just 8.8.8.8...

@SukerYan, can you ping 8.8.8.8 from china?

@Pant thank you for your PR!

SukerYan:
no, i can not ping 8.8.8.8 from China. :(

2018年2月5日 20:24，"nachoparker" <notifications@github.com>写道：

nachoparker:
can you ping 4.2.2.1 or 4.2.2.2?

SukerYan:
they are fine. thanks.

2018年2月6日 01:19，"nachoparker" <notifications@github.com>写道：

nachoparker:
applied in `v0.46.1`

nachoparker:
I'll reopen and implement it a bit better, but just relying on the return value of the clone operation

nachoparker:
Implemented in `v0.46.4`

Issue #320, opened 2017-12-07
no update - no internet connectivity


colognaut:
wanted to update from nextcloudpi 0.39 to newest version.
it always answers "no internet connectivity". 
i can ping ownyourbits.com. further updates had no problem.

Tercus:
I am currently having the same error. Also mentioned it in [this forum](https://help.nextcloud.com/t/nextcloudpi-update-fails-no-internet-connectivity/24469) thread over at the nextcloud forum.

svankan:
I have the same problem because I am not able to ping github.com. The script verify internet connection in ncp-update with this.

`ping  -W 2 -w 1 -q github.com &>/dev/null || { echo "No internet connectivity"; exit 1; }`

A friend with the same ISP have the same problem and I it works with other ISP vendors. 
I have opened a support case for that.
Is it possible change to curl instead of ping? 
As a workaround I change github.com to something else where ICMP is allowed and is working.
After the upgrade I have to change manually again.

colognaut:
@svankan where can i find the ncp-update script?

svankan:
@colognaut Look in /usr/local/bin/ directory.

colognaut:
@svankan helped. thx!

Daniel-Krebs:
I had the same problem yesterday, for me changing the DNS Server in my router did it. My ISP somehow gives wrong information for github.com (and only for github.com)

xedoc64:
Got the same problem. I also get no ICMP response from different providers, maybe github changed some network stuff at their site.

It would be nice if the host which is used for the internet connectivity check can be changed atone place (config menu or a variable in library.sh?).

ChriKn:
May I believe some of you are on untiymedia in germany ? Same problem here... 
And I can't change my DNS inside my router sadly...

Daniel-Krebs:
Germany but Vodafone

xedoc64:
Germany and Telecolumbus.

matzel82:
I have the same Problem.
I dont have any solution.

Daniel-Krebs:
@matzel82 can you change your DNS server in your router?

matzel82:
@Daniel...

Yes...its possible. How ?

Es wäre möglich. Ich nutze eine Fritzbox. In was sollte ich es ändern?

matzel82:
Its working very fine 

8.8.8.8
8.8.4.4

Daniel-Krebs:
[...]

PR #389, opened 2018-01-31
Change ping site to nextcloud.com (WAN connectivity check).

Pant:
The connectivity check to WAN that was taking place in these 6 scripts was pinging to 'google.com' (5) and 'github.com' (1). github.com was having some issues in the past and I remember the change to google.com. Although in the issue #365 it is reported that google.com is blocked from China. So I changed it to 'nextcloud.com' which based on [this site](https://www.comparitech.com/privacy-security-tools/blockedinchina/) is not blocked in China.

I would agree to change 'nextcloud.com' to anything else if anyone finds it necessary.

### DIFF
--- a/bin/ncp-check-version
+++ b/bin/ncp-check-version
@@ -4,7 +4,7 @@
 
 [ $(id -u) -ne 0 ] && exit 1
 
-ping  -W 2 -w 1 -q google.com &>/dev/null || { echo "No internet connectivity"; exit 1; }
+ping  -W 2 -w 1 -q nextcloud.com &>/dev/null || { echo "No internet connectivity"; exit 1; }
 
 rm -rf /tmp/ncp-check-tmp
 
--- a/bin/ncp-diag
+++ b/bin/ncp-diag
@@ -51,7 +51,7 @@ echo "Redis service|$( pgrep -c redis-server &>/dev/null && echo up || echo down
 echo "Postfix service|$( postfix status &>/dev/null && echo up || echo down )"
 
 # WAN
-echo "internet check|$( ping -W 2 -w 1 -q github.com &>/dev/null && echo ok || echo no } )"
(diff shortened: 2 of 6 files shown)


testbird:
Shouldn't most tests try the servers required for the next step in the scripts, and fail with something like "could not connect to $SERVER" (or whatever error message the command returns).

nachoparker:
@testbird that's what I was doing originally... I was checking github prior to making a clone from github, but it seems like github blocks ping from some countries, like Germany... this was reported and changed the ping to google for that reason 

See https://github.com/nextcloud/nextcloudpi/issues/320

@Pant I want to wait to see if 8.8.8.8 is blocked in china, I like the idea of pinging an IP better than a domain, thanks!

testbird:
>want to wait to see if 8.8.8.8 is blocked

That's an activity tracking IP you should find in every block list.


Most of the time ping is probably not the required protocol, and if you just use the wget or curl error messages
it helps to identify the required connections: https://github.com/nextcloud/server/issues/8147

Where not available, even just check $? after a bash check like this:
echo -e "GET http://ownyourbits.com HTTP/1.0\n\n" | nc ownyourbits.com 80 > /dev/null 2>&1
https://stackoverflow.com/a/26820300

testbird:
Please reopen. The ping test results for a random third party IP are not valid test if downloading the updates etc. is possible. (Simlpy attempt the requested action, and fail decently.)

And it not justified to require admisn to whitelist request to third party server in the firewall rules, and thus open it for incoming packages to be analysed, just to be able to do a download from somewhere else.

nachoparker:
Hi @testbird 

Yes, I guess we can just rely on the return value of `git clone`

You wrote me many messages when I was away... I'll read them and answer them as I have time, thanks for the feedback

Issue #125, opened 2017-08-18
ncp-web: confirmation dialog for shutdown


deljones:
Hi
Love the 0.23.0 upgrade and the continuing brilliant work on nextcloudpi....

However the shutdown button top right caught me off guard. Great additon but could do with a drop down for 'reboot', sleep, shutdown or confirmation box of the shutdown.

Dj

nachoparker:
Hi Del,

Yes, this is the first preliminar version of the feature, we need a confirmation dialog, see #119

theCalcaholic:
Oh, sorry @nachoparker . I didn't see that you had assigned this issue to yourself already. Feel free to ignore my pull request if you already have worked on it.

nachoparker:
not at all, I haven't worked on it. Thanks for the PR

nachoparker:
Implemented in #403

PR #403, opened 2018-02-10
Feature/shutdown dialog

theCalcaholic:
When pressing the power button in ncp-web, show a menu with options 'shut down' and 'reboot' (compare #125).

### DIFF
--- a/ncp-web/index.php
+++ b/ncp-web/index.php
@@ -61,7 +61,7 @@
   {
     echo <<<HTML
     <div id="first-run-wizard">
-      <div>
+      <div class='dialog'>
         <br>
         <h2 id="config-box-title">NextCloudPi First Run</h2>
         <p>Click to start the configuration wizard</p>
@@ -166,6 +166,14 @@ HTML;
           <textarea readonly id="details-box" rows="25" cols="60"></textarea>
         </div>
       </div>
+
+      <div id="poweroff-dialog" class='dialog primary' style='display:none;'>
+        <input type='submit' id='poweroff-option_shutdown' class='primary' value='shut down' />
+        <input type='submit' id='poweroff-option_reboot' class='primary' value='reboot' />
+        <input type='submit' id='poweroff-option_cancel' class='primary' value='cancel' />
+      </div>
+
+      <div id='overlay' style="display:none;"></div>
 
   </div>
 
--- a/ncp-web/index.php
+++ b/ncp-web/index.php
@@ -167,15 +167,14 @@ HTML;
         </div>
       </div>
 
-      <div id="poweroff-dialog" class='dialog primary' style='display:none;'>
-        <input type='submit' id='poweroff-option_shutdown' class='primary' value='shut down' />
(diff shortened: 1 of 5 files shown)


nachoparker:
great! I'll look at it soon

theCalcaholic:
Could you test it once (just see if shut down and reboot work as intended)? I don't completely trust my docker environment and have no NCP device at hand atm.

nachoparker:
sure, I test everything before merging it

nextcloudpi.sh
```diff
@@ -120,7 +120,7 @@ sleep 0.5 && echo "" > /run/ncp.log
 exit $RET
 EOF
   chmod 700 /home/www/ncp-launcher.sh
-  echo "www-data ALL = NOPASSWD: /home/www/ncp-launcher.sh , /sbin/halt" >> /etc/sudoers
+  echo "www-data ALL = NOPASSWD: /home/www/ncp-launcher.sh , /sbin/halt, /sbin/reboot" >> /etc/sudoers
```
nachoparker: this needs to be included too for old images, so it should be duplicated in `update.sh`

ncp-web/index.php
```diff
@@ -169,6 +169,13 @@
 
   </div>
 
+  <div id="poweroff-dialog" class='dialog primary' style='display:none;'>
```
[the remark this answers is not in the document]
theCalcaholic: That's how the minifiedjs/jquery methods do it. If we do it like this it's compatible with `$(element).hide()` and 
`.show()`.

nextcloudpi.sh
theCalcaholic: Ah right. I thought I had already added this. :+1:

nachoparker: you added it in nextcloudpi.sh which is only used for building new images. If we are to update currently running images with the new feature through `ncp-update`, then we need to do this in `update.sh`

other changes, like any changes to files in `bin`, `ncp-web` and `etc` are overwritten automatically without more intervention, but this is `/etc/sudoers` so we have to do it explicitly.

ncp-web/index.php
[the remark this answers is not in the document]
nachoparker: just use the `hidden class` and get rid of the inline style ;)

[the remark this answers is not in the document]
theCalcaholic: And that works with said functions? Oo
Alright then.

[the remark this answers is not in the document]
nachoparker: like here
https://github.com/nextcloud/nextcloudpi/blob/master/ncp-web/index.php#L157

nextcloudpi.sh
theCalcaholic: Yes, I thought I had added it twice. ;)
But apparently not so.

nachoparker: :D looking great, once those two little things are fixed, I will build a devel image with `build_devel.sh` and test it.

I want to make a command to update from devel, so we don't need to use QEMU and generate images to test.

nachoparker:
pushing it to devel for the final testing

thank you!

Issue #402, opened 2018-02-09
LetsEncrypt Certs not available after docker-compose up / down


Aeon512:
Hi, not sure if I'm doing something wrong but lets start from scratch

1. `docker-compose -f <docker-compose-armhf.yml> up -d `
2. access ncp and enable lets encrypt
3. access via https and see that its works
4. `docker-compose -f <docker-compose-armhf.yml> down `
5. `docker-compose -f <docker-compose-armhf.yml> up -d `

When looking at `dockers logs` I see a PHP stack-trace but the actual error at the very beginning is an error in Apache loading the configuration since the `/etc/letsencrypt/live` folder with the certificates no longer exists.

As far as I can tell, the step 2) creates the necessary certificate data in `/etc/letsencrypt/live` and `/etc/letsencrypt/archive` but these folders are not saved via the library functions in the permanent `/data/etc` folder.

Wouldn't it make more sense to save this folder permanently as well?

nachoparker:
good catch.

you have to add a line to the `000ncp` file

Aeon512:
Sure I'm here to help :-)
But isn't the `000ncp` file also part of the docker-image and hence changes to it are not persistent?

nachoparker:
you are absolutely right, the code is there (didn't remember), but what we need is to activate it when launching _letsencrypt_

Aeon512:
I just couldn't find anything in the early startup phase where you check for an active letsencrypt configuration

in particular since the `is_active()` method in the file mentioned above just tests for the existence of `/etc/letsencrypt/live` which yields an hen-egg problem :-)

If you have something which I should test, or if I can assist otherwise, let me know.

nachoparker:
you are right... the problem is that we need to persist the whole `/etc/services-enabled` folder ( with the `_persistent_cfg_()` trick that you have already seen, because otherwise, once you remove the container it won't remember what services should be enabled

if you feel like playing with this you can add that to the _debian-ncp_ container (or wherever it's best, talking from memory) and rebuild the container. It should work.

there's instructions on ownyourbits.com, basically you do

```
make nextcloudpi-x86
```
and that will rebuild with the changes

Aeon512:
I would vote to add it in `nextcloudpi/docker-armhf/debian-ncp/run-parts.sh` and persist the cfg in the enabled / disabled case as well as just before iterating through the enabled services.

[...]

PR #407, opened 2018-02-11
Restore LetsEncrypt after docker-compose up/down

Aeon512:
Fix for #402

### DIFF
--- a/etc/nextcloudpi-config.d/letsencrypt.sh
+++ b/etc/nextcloudpi-config.d/letsencrypt.sh
@@ -42,15 +42,23 @@ install()
   /etc/letsencrypt/letsencrypt-auto --help # do not actually run certbot, only install packages
 
   [[ "$DOCKERBUILD" == 1 ]] && {
-    cat > /etc/services-available.d/100letsencrypt <<EOF
+    # execute before lamp stack
+    cat > /etc/services-available.d/009letsencrypt <<EOF
 #!/bin/bash
 
 source /usr/local/etc/library.sh
 persistent_cfg /etc/letsencrypt
 
+cat > /etc/cron.weekly/letsencrypt-ncp <<EOFCRON
+#!/bin/bash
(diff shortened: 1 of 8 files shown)


nachoparker:
thanks! code looks good. Building now

etc/nextcloudpi-config.d/letsencrypt.sh
```diff
@@ -82,6 +90,10 @@ EOF
     # delayed in bg so it does not kill the connection, and we get AJAX response
     bash -c "sleep 2 && service apache2 reload" &>/dev/null &
     rm -rf $NCDIR/.well-known
+    
+    # update configuration
+    /usr/local/sbin/update-rc.d letsencrypt enable
```
nachoparker: a little change I made

```
[[ "$DOCKERBUILD" == 1 ]] && update-rc.d letsencrypt enable
```

, so it only gets called in the docker case

Aeon512: No, that's wrong.
The `configure()` get's called when you enable letsencrypt on the web interface.
During build only the `install()` will be called.

With this change, you can never enable letsencrypt!

nachoparker: that variable is `1`, because I used `ENV` in the dockerfile.

you can open a shell with `docker exec -ti bash nextcloudpi` and verify it

nachoparker:
merging it manually to devel, will push in a bit

the docker version needs a lot of love, and we need a lot of help! 

See this

https://github.com/nextcloud/nextcloudpi/issues?q=is%3Aissue+is%3Aopen+label%3Adocker

Thanks!

etc/nextcloudpi-config.d/letsencrypt.sh
Aeon512: You are totally right. I missed that.
You're proposal works fine. Thanks for catching that.

PR #415, opened 2018-02-15
random password provisioning on boot/startup

Aeon512:
During bootup of a new docker image, the redis
password might not match the nextcloud configuration.
Hence, we automatically update the nextcloud configuration.

**Note:** The code is taken from `nc-restore.sh` but not extensively tested.


nachoparker:
thank you again for this.

we need to do this as well for the database, so I suggest that we write a "provisioning script" that

- checks `/root/.my.cnf`, if no password set a random one there
- checks `/etc/redis/redis.conf`, if no password set a random one there
- update `config.php` with those two passwords

this script would be executed as part of the initialization both in docker and in the SD card version. In the future, we can also set the number of PHP threads automatically, because right now it is only taylored to the CPU of the pi, and we now support the odroid, and also x86 dockers.

this would also improve security because every different user would have a completely different random password.

I have been meaning to do this for a long time but my hands are pretty full right now, so I would like to suggest that we do this now and fix all those problems at once.

Would you like to help with this?

Aeon512:
Are you sure about the database?
In my understanding the database password is stored in the database and this database is already persistent. 

Of course the contents of the .my.cnf is not correct but that does not affect functionality. 

In my tests, after fixing the redis password I could successfully boot up and had a working Nextcloud

nachoparker:
you are right... and I am right too

the password in `.my.cnf` is to access the database externally as root ( with the command `mysql -u root -p password` ).

if `.my.cnf` exists, you can omit the `-p password` parameter as the password will be automatically read from that file.

So it would be a much needed security improvement to randomize that password even more.

Then we have the admin password for the nextcloud tables of the database, that is the one in `config.php` and can be changed by SQL statements ( [example](https://github.com/nextcloud/nextcloudpi/blob/master/etc/nextcloudpi-config.d/nc-nextcloud.sh#L179) ). That's the one you are talking about.

That one should also be randomized for each user for security.

I hope that makes more sense now

Aeon512:
So you mean there will be a script which automatically sets the redis password during bootup to a random value?
That’s possible and makes sense. 

For the database however, we would need to check if there is already a password defined and then reuse this one. 

Creating a new database password is not possible since you can’t access the database to change it. 

Did I get it correctly?

nachoparker:
yes, we are almost on the same page

I want to create a script that *if the system is uninitialized* initializes it to a unique random password scheme.

So I would only change the redis password if it doesn't exist currently, or if it doesn't match with the one on config.php.

Same with the database passwords ( the nc-admin password and the mysql password ): I would check if it exists, and if it doesn't or it doesn't match I would update it with a random value everywhere ( config.php and database itself through SQL ).

Then, we would ship images and docker containers in this "uninitialized state", so during the first run we "provision random password" for the services. This would improve security.

> Creating a new database password is not possible since you can’t access the database to change it.

We would do it through SQL, as in [here](https://github.com/nextcloud/nextcloudpi/blob/master/etc/nextcloudpi-config.d/nc-nextcloud.sh#L179). This works as long as we have a valid mysql password in `.my.cnf`

Thoughts?

Aeon512:
For the reds issue I tried to set it up according to your suggestion.
However, it's not yet tested carefully - just as a first step to see if we are on the same page

nachoparker:
your commit looks great. I would just put that code in some script, for instance `/usr/local/bin/ncp-provisioning` and run that script from both `020nextcloud` and in the case of the non docker image from a systemd service (we can do that last).

Then, do the same thing for the database.

Aeon512:
I moved the code to the `ncp-provising.sh` file but also consolidate the identical files in `docker` and `docker-armhf` to a single `docker-common` to avoid duplicated code.

Note that I didn't have the time to actually test this change. Docker is building but it will be busy the next days, so testing might take a while. Nevertheless wanted to share this directly.

Aeon512:
I tested and correct some minor issues.
From my side it would be ready for integration.

nachoparker:
Looks great, I'll test it soon

> I moved the code to the ncp-provising.sh file but also consolidate the identical files in docker and docker-armhf to a single docker-common to avoid duplicated code.

Yes, I welcome that :)

nachoparker:
by the way, in case you are interested, we have a `@NextCloudPi` public group on Telegram. Originally it was dev only but now is open for everyone.

In case you want to be in tune with what's happening and help with opinions and such we are typically more responsive there.

nachoparker:
one little thing: please in the future send PRs over the `devel` branch, as explained [here](https://github.com/nextcloud/nextcloudpi/wiki/Workflow-and-testing-of-new-features)

docker-common/nextcloud/ncp-provisioning.sh
```diff
@@ -0,0 +1,34 @@
...
+DROP USER '$DBADMIN'@'localhost';
+CREATE USER '$DBADMIN'@'localhost' IDENTIFIED BY '$DBPASSWD';
+GRANT ALL PRIVILEGES ON nextcloud.* TO $DBADMIN@localhost;
+EXIT
+EOF
+}
+test -f /data/app/config/config.php && {
+  echo Updating NextCloud config with MariaDB password $DBPASSWD
+  sed -i "s|'dbpassword' =>.*|'dbpassword' => '$DBPASSWD',|" /data/app/config/config.php
+}
```
nachoparker: I am adding an exit 0 here, because in the first run, the `test` for `config.php` fails, then the provisioning script returns error, and because of that redis and nc-init never run on `020-nextcloud`

nachoparker:
it works, thank you very much.

the next step is to use this for the SD card image version as well

PR #427, opened 2018-02-23
Added some useful comments for first time users

ovpc:
added:
```
INFOTITLE="SSH notes"
INFO="In order to enable SSH, the password for user pi can NOT remain set to the default raspberry. 
You HAVE to create a NEW password for pi if you want this program to enable SSH, it will fail if you dont!
Note: Use normal AlphaNumeric, the only special characters allowed are .,@-_/"
```


nachoparker:
thanks, will apply now. Little reminder: next time, we should send PRs to the `devel` branch

PR #499, opened 2018-05-09
Improve tempdir handling

ThomasKaiser:
No need to introduce temp file vulnerabilities. I hope other scripts are not also affected and only checked this script since https://github.com/armbian/config/commit/b6d2249dff8df8826cb3280255aedf9919f5dff4#commitcomment-28910611

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -14,7 +14,8 @@ BRANCH=master
 
 set -e$DBG
 
-TMPDIR=/tmp/nextcloudplus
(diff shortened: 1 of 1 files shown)


nachoparker:
thanks Thomas, for whatever reason, didn't get notified of this comment https://github.com/armbian/config/commit/b6d2249dff8df8826cb3280255aedf9919f5dff4#commitcomment-28910611

PR #567, opened 2018-07-20
Update some Chinese translations

chiyi4488:
Update some Chinese translations


nachoparker:
how about this PR? are you finished with it? should I apply it too?

nachoparker:
I also updated this. Again, let's see if it is all good by running `sudo ncp-update devel`.

Many thanks!!!

PR #568, opened 2018-07-20
Modify all zn.json file names to zh_TW.json And update some Chinese translations

chiyi4488:
Modify all zn.json file names to zh_TW.json
And update some Chinese translations


chiyi4488:
Ok, because no one will use zn as the code name for Chinese translation. Usually we will use the zh-region because I am from Taiwan, so I use TW.

nachoparker:
I see, well the thing is that ncp-web will try to automatically set the language based on the browser language configuration `HTTP_ACCEPT_LANGUAGE`. I would like to know which one exactly you are seeing in your browser, because that is what we should use. I thought it was `zn` but I might be wrong.

chiyi4488:
I found my name is wrong, it should be zh-TW

chiyi4488:
The way I name it is named after the nextcloud naming method.

nachoparker:
we can try something. Create a file named test.php in `/var/www/ncp-web` that looks like this

```
<?php var_dump($_SERVER['HTTP_ACCEPT_LANGUAGE']); ?>
```

And then, open it from your browser `https://<nextcloudpi>:4443/test.php`.

What does it say?

chiyi4488:
string(35) "zh-TW,zh;q=0.9,en-US;q=0.8,en;q=0.7"

chiyi4488:
So should I rename it?

nachoparker:
thanks! that worked... yes, it seems like the right name would be `zh`, or `zh-TW`. The first one is more generic and I think it is better.

For example, we do `es` for Spanish, not `es-ES`. I do not know what the implications are for Chinese, but please chose the one that is more general to allow more people to understand it if possible.

You can just keep modifying your github fork and the changes will be reflected here. You don't need to open a new PR.

Ask me anything you need!

Many thanks :)))

chiyi4488:
Also, I think I can do another Chinese in Nextcloud. Because Chinese is actually divided into two types, one is Traditional Chinese (I am doing it now) and the other is Simplified Chinese. However, Simplified Chinese is very simple to make. Just translate my traditional Chinese into Simplified Chinese using Google Translate.

However, I must first translate the remaining traditional Chinese.

nachoparker:
That is great Yi! so simplified chinese would be `zh`? `zh-XXX`?

Tell me when you are done with Traditional Chinese and we can test it.

Thanks

chiyi4488:
zh-cn

chiyi4488:
So, can you please pass my application?

nachoparker:
ok, if you are finished we will test your changes

nachoparker:
I will explain later. 

What about your other PR #567?

Why don't you put all your changes from #567 here in #568 and I will apply everything together

PR #572, opened 2018-07-22
Fixes #566 Remove redundant opcache configuration

### LINKED ISSUE #566: Cannot Load Zend OPcache - its already loaded
I have noticed with the release of v0.57.18 that I am seeing this error in multiple locations:

Cannot load Zend OPcache - it was already loaded

Thanks

Iolaum:
PR sent following our discussion in #566

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -289,6 +289,10 @@ EOF
   install_script nc-backup.sh
   cd -          &>/dev/null
 
+  # remove redundant opcache configuration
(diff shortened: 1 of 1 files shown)


nachoparker:
Same fix again... I guess we will leave this until they fix [the bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=815968)

Thanks!!!

PR #594, opened 2018-08-18
removed trailing - from -O option in wget

aj0221:


aj0221:
trying to install this on my pi this morning and got a 404; the trailing - was messing with the pipes/wget

aj0221:
Ah, my bad. Made a mistake. This is not correct.

PR #602, opened 2018-08-22
Change order in configuration entries

TomTurnschuh:
The email-entry is moved below the option to activate email notifications.


nachoparker:
thanks, applied

PR #605, opened 2018-08-23
Set pretty URLs

enoch85:
In the Nextcloud VM we do this during the installation as well to avoid the index.php path even when not using LE.


enoch85:
Please SQUASH and merge after review

nachoparker:
thanks for this @enoch85 

I had this debate with myself some time ago. I decided not to include pretty URLs just to make it as fast as possible. I also disabled some apache modules back in the day for the same reason.

But I ask myself every now and then if this is the right decision. Definitely nice to have nicer URLs. 

On another slightly unrelated topic, I have been thinking lately to disable htaccess, as it is another way to improve performance.

Thoughts?

enoch85:
> On another slightly unrelated topic, I have been thinking lately to disable htaccess, as it is another way to improve performance.

I don't think that's a good idea as many of the security "patches" happens in .htaccess. In the VM it's not used per say as PHP-FPM won't read .htaccess but we do copy the settings to .user.ini so it has the same effect anyway.

Or, do you know a way to benefit from the settings in .htaccess and at the same time boost performance?

enoch85:
> Regarding pretty urls, folks in the Telegram group are suggesting making it a ncp option in ncp-web. Maybe it's the best solution

Why would you want `index.php` in your URL? xD

> maybe take those settings to the virtual hosts file or similar.

That's what user.ini are for. We have an existing function in the VM already. :)

nachoparker:
> That's what user.ini are for. We have an existing function in the VM already. :)

do you mean the `.user.ini` file in the NC folder? I am guessing (just guessing) that it suffers from the same problem than `.htaccess`, they need to be accessed every time there is a web request.

> Why would you want index.php in your URL? xD

Only to squeeze some performance (save the mod_rewrite execution time), on a rpi2 or rpi zerow everything counts, so that is why I didn't implement pretty urls in the first place. It could be optional though, so the user can activate it if he has a bulkier computer.

enoch85:
> Only to squeeze some performance (save the mod_rewrite execution time), on a rpi2 or rpi zerow everything counts, so that is why I didn't implement pretty urls in the first place. It could be optional though, so the user can activate it if he has a bulkier computer.

Sure, we're on rPi, didn't think of that. 

I think it's a good idea to make it optional. :)

enoch85:
Closing this as it will be optional instead. Please reopen otherwise.

Thanks!

nachoparker:
Let's keep this open, so I don't forget ;)

nachoparker:
Implemented as an app in #625 

Thanks!

PR #607, opened 2018-08-23
Add Referrer-Policy

TomTurnschuh:
With this addition, a security warning in Nextcloud 14 can be prevented.


TomTurnschuh:
Additional headers like the X-Content-Type-Options or the X-Frame-Options could be worth including as well.

TomTurnschuh:
> Additional headers like the X-Content-Type-Options or the X-Frame-Options could be worth including as well.

Well, I guess they are already set by Nextcloud itself.

nachoparker:
Great, thanks! I will test it together with the jump to NC14

PR #624, opened 2018-09-07
Change email program from PHP to Sendmail

TomTurnschuh:
PHP seems to be no longer supported in Nextcloud 14.
This modification might need further testing in different environments.


nachoparker:
Hi, thanks!

have you tested this change? you can send a test email if you have your email details configured in the user section.

TomTurnschuh:
I tested it on my ODROID and Raspberry Pi devices, but I cannot conduct tests in different environments.
I only changed the config file on my instances, and did not test with the altered nc-init file. However, the new value corresponds to the config entry when setting this directly in Nextcloud.

nachoparker:
So you sent and received the test email... great.

Yeah if you made the change through that command and that works don't need to test the nc-init part.

Thanks again!

nachoparker:
Thank you, applied.

I modified your commit to include the modification in `update.sh` so that users that upgrade get the change.

Thanks!

PR #631, opened 2018-09-14
Update nc-format-USB.sh

### LINKED ISSUE #630: Issue with nc-format-USB
There is an issue with formatting a mounted drive:

1. Install and use NCP and NC, with data on a hard drive
2. Reinstall NCP
3. Enabling automount
4. The hard drive cannot be formated even though it is indeed mounted. The error says:
`ERROR: counted 0 devices. Please, only plug in the USB drive you want to format`

I didn't test yet with a unformated drive.

TomTurnschuh:
Resolves #630

Delete `-1` in calculation of drives number.


nachoparker:
thanks, good catch

PR #632, opened 2018-09-15
DDNS_spDYN: Add option for IPv6

### LINKED ISSUE #615: spdyn.de  IPv6
I had problems with spdyn.de....
Found out it had to with IPv6 vs IPv4.

In nextcloudpi/etc/ncp-config.d/spDYN.sh "https://api.ipify.org/" is used to get the ip adress. But this api only provides IPv4.

I guess for the future you should use an API wich provides IPv6 (like "https://myexternalip.com/raw", wich i am using now:)  )

TomTurnschuh:
Closes #615.

Adds the ability to use IPv6 addresses with spDYN.

@nachoparker: For a NCP update the following steps are required:
1. Delete folder (including content) `/usr/local/etc/spdnsupdater`
OR delete file `/usr/local/etc/spdnsupdater/spdnsUpdater.sh` (which is the content of above folder)
2. Run installation of `DDNS_spDYN`


TomTurnschuh:
Maybe we shoud wait with this extension. It seems that letsencrypt does not support IPv6 addresses. Is that true?

nachoparker:
@TomTurnschuh it seems like it does support ipv6

https://letsencrypt.org/2016/07/26/full-ipv6-support.html

etc/ncp-config.d/DDNS_spDYN.sh
```diff
@@ -122,11 +128,11 @@ configure()
     
     # Adds file to cron to run script for DNS record updates and change permissions 
     touch $CRONFILE
-    echo "0 * * * * root $INSTALLPATH/spdnsUpdater.sh $DOMAIN_ $TOKEN_ >/dev/null 2>&1" > "$CRONFILE"
+    echo "0 * * * * root $INSTALLPATH/spdnsUpdater.sh $DOMAIN_ $TOKEN_ $IPv6 >/dev/null 2>&1" > "$CRONFILE"
```
nachoparker: shouldn't this be `$IPv6_`?

etc/ncp-config.d/DDNS_spDYN.sh
```diff
@@ -122,11 +128,11 @@ configure()
...
     chmod +x "$CRONFILE"
 
     # First-time execution of update script and print response from spdns.de server
-    "$INSTALLPATH"/spdnsUpdater.sh "$DOMAIN_" "$TOKEN_"
+    "$INSTALLPATH"/spdnsUpdater.sh "$DOMAIN_" "$TOKEN_" "IPv6"
```
nachoparker: shouldn't this be `$IPv6_`?

etc/ncp-config.d/DDNS_spDYN.sh
TomTurnschuh: Thats true.
I did the changes on my device and in Github in parallel – and sometimes forgot to update on both channels.
Thanks for correcting.

TomTurnschuh:
> @TomTurnschuh it seems like it does support ipv6

That's true for the Let's Encrypt side, but somehow it does not work on my device, even though the cloud itself is accessible with a browser.
I think, maybe the certbot package instead of letsencrypt should be used.

The response is:
```
To fix these errors, please make sure that your domain name was
entered correctly and the DNS A record(s) for that domain
contain(s) the right IP address.
...
```
That's why I thought this could be the problem. The same settings with an IPv4 address works well.

PR #634, opened 2018-09-16
Two variable fixes in DDNS_spDYN.sh

TomTurnschuh:
With these fixes it does work correctly with both IPv4 and IPv6 on my installation.


TomTurnschuh:
For discussion: spDYN has its own URLs to get IP addresses in raw format, however, they don't support encryption but only HTTP. Is it still worth to use them, as the can be assumed to work as long as their DDNS service is online?

TomTurnschuh:
Related discussion in #635.

nachoparker:
Thank you!

PR #637, opened 2018-09-17
Don't overwrite an existing mail_smtpmode if it is not "PHP"

TomTurnschuh:
Overwriting the value `SMTP` is probably not the best idea. `SMTP` settings are done manually as required. They should not be disabled or overwritten.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -145,7 +145,10 @@ EOF
   }
 
   # NC14 doesnt support php mail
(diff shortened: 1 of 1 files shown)


nachoparker:
Thanks good catch. You forgot to PR against `devel`. This time I'll just apply it manually, so that the fix goes out ASAP

TomTurnschuh:
This time I went for the master branch on purpose. :-)

PR #643, opened 2018-09-21
DDNS_spdyn.sh : Send new IP only when changed.

TomTurnschuh:
DDNS providers usually don't like updates on the IP address when there is no change.


nachoparker:
Applied, thank you!

PR #645, opened 2018-09-21
DNS serveur

Jeanoo:
Hello,
do you know if it is possible to install other functions on the nextcloudpi distribution
for example: dns server
Thank you very much.


nachoparker:
We can talk about requesting new features, but most importantly somebody needs to implement them.

In any case, we already have a DNS server

https://ownyourbits.com/2017/03/09/dnsmasq-as-dns-cache-server-for-nextcloudpi-and-raspbian/

Jeanoo:
Thank you very much.

PR #646, opened 2018-09-22
(add function)Fix some errors that nextcloud will alert after the update.

chiyi4488:
Fix some errors that nextcloud will alert after the update.

*****This is just one of my ideas and has not been tested. *****
After updating to NC14, the system will display security and other issues in the overview. If can fix the problem before the display, the user will be much more convenient.

Please ask the main developers to test, thank you.

### DIFF
--- a/bin/ncp-update-nc
+++ b/bin/ncp-update-nc
@@ -166,7 +166,7 @@ echo "Fix permissions..."
 chown -R www-data:www-data nextcloud
 find nextcloud/ -type d -exec chmod 750 {} \;
 find nextcloud/ -type f -exec chmod 640 {} \;
-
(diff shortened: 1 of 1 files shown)


nachoparker:
Awesome, thanks chiyi. I think we can put this better at the end of `bin/nc-update-nc`.

chiyi4488:
Put on "clean up"??

chiyi4488:
Or "done"??

nachoparker:
I would say... [here](https://github.com/nextcloud/nextcloudpi/blob/master/bin/ncp-update-nc#L174)

chiyi4488:
I change it. look like ok?

nachoparker:
great, thank you chiyi!

PR #654, opened 2018-09-28
nc-prettyURL: Catch failure of maintenance:update:htaccess

TomTurnschuh:
Show error in case of failure of maintenance:update:htaccess.

#### TODO
- [ ] The original app state should be regained in case of failure.

### DIFF
--- a/etc/ncp-config.d/nc-prettyURL.sh
+++ b/etc/ncp-config.d/nc-prettyURL.sh
@@ -21,10 +21,20 @@ configure()
   [[ $ACTIVE_ != "yes" ]] && {
     sudo -u www-data php "$OCC" config:system:set htaccess.RewriteBase --value=""
     sudo -u www-data php "$OCC" maintenance:update:htaccess
+    [[ $? -ne 0 ]] && {
+      echo "There has been an error."
+      # ACTIVE_ = "yes" (TODO)
+      return 1
+    }
     echo "Your cloud does no longer have a pretty domain name."
   } || {
     sudo -u www-data php "$OCC" config:system:set htaccess.RewriteBase --value="/"
     sudo -u www-data php "$OCC" maintenance:update:htaccess
+    [[ $? -ne 0 ]] && {
+      echo "There has been an error."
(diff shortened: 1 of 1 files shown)


etc/ncp-config.d/nc-prettyURL.sh
```diff
@@ -21,10 +21,20 @@ configure()
   [[ $ACTIVE_ != "yes" ]] && {
     sudo -u www-data php "$OCC" config:system:set htaccess.RewriteBase --value=""
     sudo -u www-data php "$OCC" maintenance:update:htaccess
+    [[ $? -ne 0 ]] && {
+      echo "There has been an error."
+      # ACTIVE_ = "yes" (TODO)
```
nachoparker: please get rid of this comment and merge/rebase to the latest devel

nachoparker:
Thanks! Looks good to me. Regarding the activation, the way to go is to include an `is_active()` step. If this function exists will override the existence of `ACTIVE_` in order to show the app as active.

[Example](https://github.com/nextcloud/nextcloudpi/blob/master/etc/ncp-config.d/nc-datadir.sh#L19)

etc/ncp-config.d/nc-prettyURL.sh
```diff
@@ -16,6 +16,13 @@ OCC="$NCDIR/occ"
 
 install() { :; }
 
+isactive()
+{
+  local REWRITEBASE
+  REWRITEBASE="$( grep RewriteBase /var/www/nextcloud/.htaccess )" || return 1
```
nachoparker: cool, looks good

so when it is disabled, `Rewritebase` disappears from `.htaccess`, right?

have you tested it?

TomTurnschuh: I couldn't test this yet. We might need to check the value, but I yet have to test it.

TomTurnschuh: `RewriteBase` is indeed deleted from `.htaccess`. Hence, the function is correct.

Issue #470, opened 2018-04-07
Preview Generator App Feedback on NCP @ Rpi2


Iolaum:
I 've enabled the [Preview Generator App](https://apps.nextcloud.com/apps/previewgenerator) and have been testing it on my raspberry pi 2 NCP installation.

**What's the problem?**

Generating preview images is too hard from raspberry pi's and therefore performance downgrades when viewing a folder with images.

**Possible solution:**
Installing the Preview Generator on the raspberry pi allows the creation of those preview images ahead of time so the rpi doesn't have to compute them when you browse.

Unfortunately depending on your usage this may not be straightforward. On a load of about 135 Gb of photos my rpi2 has taken two weeks to generate around 2/3 of previews needed ... :( 

This issue is about exploring the issue, considering how we can enable Preview Generator on NCP and some tips and tricks on people who want to enable it on their own.

Steps to enable Preview Generator  on your own.

1. Go to the apps page of your local NextCloud instance and enable Preview Generator.
2. Log into your your machine hosting your NCP instance, preferable through SSH.
3. The commands available from the app after it is installed are listed in the [projects README page](https://github.com/rullzer/previewgenerator/#commands). The first things you need to do is generate a preview of all existing images and the set up a periodic run to generate previews for new images.

Because the generation of the preview images takes ages !! you need to to it in a slightly special way by using the `screen` command. It enables you to create a session that will continue to run even after you log out - because you don't want to have to stay logged in for 1 week/50GbPhotos. 

More specifically for the raspberry pi image of NextCloudPi:

```
# become root user
$ sudo -i
# create a new screen 
$ screen
# Click enter if needed after the interactive information that pops up
# Go to the Nextcloud directory
$ cd /var/www/nextcloud
# Start the preview generation process:
sudo -u www-data php ./occ preview:generate-all
# Get out of the screen session:
Ctrl+A+D # Just hit those 3 keys together and you get detached from the screen
```

Now your rpi has started creating the preview images you may want to see what's it doing. Here's how to do it:

```
# First log in and get a terminal up
# Use top to see the running process
$ top # php having 100% usage is normal
# become root
$ sudo -i
# See available screen sessions
$ screen -list
# Assuming you have only one screen session running, connect to it
$ screen -r
# You can now see how fat the generate-all command has proceeded.
Ctrl+a+d # detach from session to let it keep running
# Lastly you may want to monitor your rpi temperature - result are in mC
[...]

PR #662, opened 2018-09-29
Create nc-previews-auto.sh

TomTurnschuh:
:no_entry: Do not merge yet :no_entry:

### TODO
- [x] Install and enable the app
- [x] Test kill feature
- [x] Verify time calculations
- [x] Test SMALLONLY
- [x] Define generated SMALLONLY sizes (which ones are often needed?)
- [x] Create `isactive()`

The following mechanisms should first be implemented in the previewgenerator itself before merging this very app into live NCP:
- [ ] Previews of deleted files must be deleted (https://github.com/nextcloud/nextcloudpi/pull/662#issuecomment-437489535)
- [ ] Preview database should update to all existing files (rullzer/previewgenerator#134)
- [ ] Preview database should update to new settings (rullzer/previewgenerator#134)
- [ ] Preview generation must be stopped before storage is entirely full

### DIFF
--- a/etc/ncp-config.d/nc-previews-auto.sh
+++ b/etc/ncp-config.d/nc-previews-auto.sh
@@ -32,10 +32,10 @@ configure()
     STARTMIN=$((  $STARTTIME_ % 60 ))
     STOPHOUR=$(( ($STARTTIME_ + RUNTIME_) / 60 ))
     STOPHOUR=$((  $STOPHOUR   % 24 ))
-    STOPMINS=$(( ($STARTTIME_ + RUNTIME_) & 60 ))
+    STOPMIN=$((  ($STARTTIME_ + RUNTIME_) & 60 ))
   
-  echo "${STARTMINS}  ${STARTHOUR}  *  *  *  root  /usr/bin/sudo -u www-data /usr/bin/php /var/www/nextcloud/occ preview:pre-generate" >  /etc/cron.d/nc-previews-auto
-  echo "${STOPMINS}   ${STOPHOUR}   *  *  *  root  /usr/bin/pkill -f \"occ preview\""
+  echo "${STARTMIN}  ${STARTHOUR}  *  *  *  root  /usr/bin/sudo -u www-data /usr/bin/php /var/www/nextcloud/occ preview:pre-generate" >  /etc/cron.d/nc-previews-auto
+  echo "${STOPMIN}   ${STOPHOUR}   *  *  *  root  /usr/bin/pkill -f \"occ preview\""
   service cron restart
 
   echo "Automatic preview generation enabled"
--- a/etc/ncp-config.d/nc-previews-auto.sh
+++ b/etc/ncp-config.d/nc-previews-auto.sh
@@ -39,8 +39,8 @@ configure()
   if [ $SMALLONLY_ == "yes" ]]
     then
       sudo -u www-data php /var/www/nextcloud/occ config:app:set --value="32"  previewgenerator squareSizes
-      sudo -u www-data php /var/www/nextcloud/occ config:app:set --value="64 128 256 512" previewgenerator widthSizes
-      sudo -u www-data php /var/www/nextcloud/occ config:app:set --value="64 128 256" previewgenerator heightSizes
+      sudo -u www-data php /var/www/nextcloud/occ config:app:set --value="128 256 512" previewgenerator widthSizes
+      sudo -u www-data php /var/www/nextcloud/occ config:app:set --value="128 256" previewgenerator heightSizes
     else
       sudo -u www-data php /var/www/nextcloud/occ config:system:delete previewgenerator squareSizes
       sudo -u www-data php /var/www/nextcloud/occ config:system:delete previewgenerator widthSizes
--- a/etc/ncp-config.d/nc-previews-auto.sh
+++ b/etc/ncp-config.d/nc-previews-auto.sh
@@ -58,7 +58,11 @@ configure()
   return 0
 }
 
-install() { :; }
+install() {
+  # install preview generator
+  sudo -u www-data php /var/www/nextcloud/occ app:install previewgenerator
+  sudo -u www-data php /var/www/nextcloud/occ app:enable previewgenerator
+  }
 
 # License
 #
--- a/etc/ncp-config.d/nc-previews-auto.sh
+++ b/etc/ncp-config.d/nc-previews-auto.sh
@@ -2,16 +2,19 @@
 
 # Periodically generate previews for the gallery
 #
-# Copyleft 2018 by Timo Stiefel
+# Copyleft 2018 by Timo Stiefel and Ignacio Nunez Hernanz <nacho _a_t_ ownyourbits _d_o_t_ com>
 # GPL licensed (see end of file) * Use at your own risk!
 #
 
 ACTIVE_=no
 STARTTIME_=240
-RUNTIME_=1
+RUNTIME_=180
 
 DESCRIPTION="Periodically generate previews for the gallery"
-INFO="Set the STARTTIME in minutes after midnight<br>and RUNTIME in minutes."
+INFO="Set the STARTTIME in minutes after midnight and RUNTIME in minutes.
+
+Make sure that you have the correct time zone set.
+You can use "sudo tzselect" in shell for that."
 
 configure()
 {
--- a/etc/ncp-config.d/nc-previews-auto.sh
+++ b/etc/ncp-config.d/nc-previews-auto.sh
@@ -9,8 +9,12 @@
 ACTIVE_=no
 STARTTIME_=210
 RUNTIME_=60
-SHOW_SMALLONLY_=yes
-GEN_SMALLONLY_=yes
+SMALLONLY_=yes
+  # Only use when not using SHOW_SMALLONLY_
+#GEN_SMALLONLY_=yes
+  # Only use in combination with SHOW_SMALLONLY_
+#SHOW_SMALLONLY_=yes
+  # Only use in combination with GEN_SMALLONLY_
 
 DESCRIPTION="Periodically generate previews for the gallery"
 INFO="Set the STARTTIME in minutes after midnight and RUNTIME in minutes.
@@ -29,8 +33,9 @@ isactive()
 configure()
 {
   [[ $ACTIVE_ != "yes" ]] && { 
-    rm /etc/cron.d/nc-previews-auto
+    rm -f /etc/cron.d/nc-previews-auto
     service cron restart
+    echo ""
     echo "Automatic preview generation disabled"
     return 0
   }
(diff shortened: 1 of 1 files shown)


nachoparker:
They can be. Unfortunately I already have an implemented this myself

```
#!/bin/bash

# Periodically generate previews for the gallery
#
# Copyleft 2018 by Ignacio Nunez Hernanz <nacho _a_t_ ownyourbits _d_o_t_ com>
# GPL licensed (see end of file) * Use at your own risk!
#
# More at nextcloudpi.com
#

ACTIVE_=no
DESCRIPTION="Generate previews periodically for the gallery"

INFO="This will make browsing the gallery much more smooth
by updating previews every night"

install()
{
  apt-get update
  apt-get install --no-install-recommends -y rsync
}

configure()
{
  echo "30  2  *  *  *  root  /usr/bin/sudo -u www-data /usr/bin/php /var/www/nextcloud/occ preview:pre-generate" >  /etc/cron.d/ncp-previews-auto
  echo "0   5  *  *  *  root  /usr/bin/pkill -f \"occ preview\""                                                  >> /etc/cron.d/ncp-previews-auto

  [[ $ACTIVE_ != "yes" ]] && { 
    rm /etc/cron.d/ncp-previews-auto
    echo "auto previews disabled"
    return 0
  }

  service cron restart
  echo "auto previews enabled"
}
```

, but I didn't push it because I didn't test it thoroughly yet.

Sorry for the misscommunication, we probably should talk before diving into code :S

nachoparker:
what do you think about this? I tried it briefly and I think it was working, but I am not sure because `ncc preview:pre-generate` didn' t print any output, as opposed to `ncc preview:generate-all`. 

The CPU was busy though so I want to think that it was working, but haven't verified

TomTurnschuh:
Hey, where is my last comment on this? :confused:

---

I personally would let the admins choose about the starting time. Depending on other tasks that are done during night it would give more flexibility.
Anyway, the presumably biggest challenge we both did not tackle yet: How can we ensure that the local time is used for the cron job and not UTC? Running it in the afternoon in Japan or Korea might not be the best idea.

**Sidenote**: Don't you want to name it `nc-previews-auto` instead of `ncp-previews-auto`?

nachoparker:
> Don't you want to name it nc-previews-auto instead of ncp-previews-auto?

Ups, totally! :D 

I guess we have to rely on them to set up the timezone properly?

TomTurnschuh:
> I guess we have to rely on them to set up the timezone properly?

This is probably the only way to go. Can we create an app for this? Or is an info enough?

nachoparker:
I would leave this to the user. Have you tried `ncc preview:pre-generate`? I did briefly but I am not sure if it is working. It didn't print out anything, but the CPU was busy. Didn't have time to look into it more.

langfingaz:
> I don't know yet where preview files are stored. However, it once seemed that after running `ncc preview:pre-generate` previews were loaded quicker. But I don't have a proof yet.

The preview files are stored in `data/appdata_<instanceid>/preview`.
For example in my case that is: `/media/myCloudDrive/ncdata/appdata_oc65v9pdw1g0/preview`

nachoparker:
Killing gracefully should be fine, if the app is well built that is. If we are doing `pre-generate` as opposed to `generate-all`, then it should not need to scan for the files because supposedly it knows what are the new files that need previews. The way I understand it is that the app writes this down in its database as new files are added.

This is to be verified

Well, this is just the beginning, more testing needs to be done.

langfingaz:
> If the cron job starts too often the previous job after the upload will not have finished.
> I think preview generator can handle running in two instances [according to the lead developer](https://github.com/rullzer/previewgenerator/issues/110#issuecomment-383475718). This covers this case quite well.

Yesterday I started to generate previews for around 1.5TB of data. As the preview:generate-all php command only runs on one thread this takes quite a while...
Therefore I started the preview:generate-all command a second time in parallel today to see if it is faster and if there might by any conflicts.

After just a minute when both processes started calculating pictures for the same folder, one of them was stopped because of the following error:

<details>
<summary>ssh terminal 1 excerpt</summary>

```
2018-10-10T12:49:24+00:00 Scanning folder /ncp/files/Archiv/Pictures/2016/09
2018-10-10T12:50:02+00:00 Scanning folder /ncp/files/Archiv/Pictures/2016/10
2018-10-10T12:52:14+00:00 
2018-10-10T12:52:14+00:00 In AbstractMySQLDriver.php line 66:
2018-10-10T12:52:14+00:00                                                                                
2018-10-10T12:52:14+00:00   An exception occurred while executing 'INSERT INTO `oc_filecache` (`mimepar  
2018-10-10T12:52:14+00:00   t`,`mimetype`,`mtime`,`size`,`etag`,`storage_mtime`,`permissions`,`parent`,  
2018-10-10T12:52:14+00:00   `checksum`,`path_hash`,`path`,`name`,`storage`) SELECT ?,?,?,?,?,?,?,?,?,?,  
2018-10-10T12:52:14+00:00   ?,?,? FROM `oc_filecache` WHERE `storage` = ? AND `path_hash` = ? HAVING CO  
2018-10-10T12:52:14+00:00   UNT(*) = 0' with params ["5", "6", 1539175934, 0, "edcdad85404533d3f96a0c0d  
2018-10-10T12:52:14+00:00   e0fc53d7", 1539175934, 27, 248407, "", "1fc6e213924d4d2ff52c24b3c56e35e6",   
2018-10-10T12:52:14+00:00   "appdata_oc65v9pdw1g0\/preview\/12959\/1365-1024.jpg", "1365-1024.jpg", 3,   
2018-10-10T12:52:14+00:00   3, "1fc6e213924d4d2ff52c24b3c56e35e6"]:                                      
2018-10-10T12:52:14+00:00                                                                                
2018-10-10T12:52:14+00:00   SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '3-1f  
2018-10-10T12:52:14+00:00   c6e213924d4d2ff52c24b3c56e35e6' for key 'fs_storage_path_hash'               
2018-10-10T12:52:14+00:00                                                                                
2018-10-10T12:52:14+00:00 
2018-10-10T12:52:14+00:00 In PDOStatement.php line 107:
2018-10-10T12:52:14+00:00                                                                                
2018-10-10T12:52:14+00:00   SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '3-1f  
2018-10-10T12:52:14+00:00   c6e213924d4d2ff52c24b3c56e35e6' for key 'fs_storage_path_hash'               
2018-10-10T12:52:14+00:00                                                                                
2018-10-10T12:52:14+00:00 
2018-10-10T12:52:14+00:00 In PDOStatement.php line 105:
2018-10-10T12:52:14+00:00                                                                                
2018-10-10T12:52:14+00:00   SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '3-1f  
2018-10-10T12:52:14+00:00   c6e213924d4d2ff52c24b3c56e35e6' for key 'fs_storage_path_hash'               
2018-10-10T12:52:14+00:00                                                                                
2018-10-10T12:52:14+00:00 
2018-10-10T12:52:14+00:00 preview:generate-all [<user_id>]
2018-10-10T12:52:14+00:00 
Done. Press any key...
```
</details>

The second process continued without any problems.

<details>
<summary>ssh terminal 2 excerpt</summary>

```
...
2018-10-10T12:44:18+00:00 Scanning folder /ncp/files/Archiv/Pictures/2016/09
2018-10-10T12:50:02+00:00 Scanning folder /ncp/files/Archiv/Pictures/2016/10
```
</details>


So it looks like you should better avoid running the preview generation multiple times at once.

nachoparker:
>  If we are doing pre-generate as opposed to generate-all, then it should not need to scan for the files because supposedly it knows what are the new files that need previews. The way I understand it is that the app writes this down in its database as new files are added.

etc/ncp-config.d/nc-previews-auto.sh
```diff
@@ -0,0 +1,87 @@
...
+  
+  # Required values for SMALLONLY previews (in pixel as used with standard DPI screens without zooming):
+  #   Thumbnail in files list: square 32
+  #   Sidebar preview (when opening "Details"): width 512; (also height 256?)
+  #   Gallery folder preview: width 128, 256; heigth 128, 256
+  #   Gallery preview: width 512; height 256
+  # 
+  if [ $SMALLONLY_ == "yes" ]
+    then
+      sudo -u www-data php /var/www/nextcloud/occ config:app:set --value="32"  previewgenerator squareSizes
```
nachoparker: looks great! good job

nachoparker: I had been thinking about adding

```
'preview_max_x' => 1024,
'preview_max_y' => 768,
'preview_max_scale_factor' => 1,n
```

to the previews app. I think that we should add this and your options either as the default choice, or even just hardcoded for NCPi users

TomTurnschuh: I wouldn't hardcode `preview_max_x` and  `preview_max_y` as they also affect the presentation of images in the gallery (not only the overview).
I personally would also not include it in SMALLONLY, for that bigger previews (eg. for the gallery) are still generated if really needed.

How about adding another option with the recommendation to enable it in lowest-end devices? Or do we then have too many optional stuff?

TomTurnschuh: I could live with hardcoding the SMALLONLY functionality, but wouldn't do it for `preview_max`.

TomTurnschuh:
Disabling SMALLONLY does not work yet. This is an issue of the `previewgenerator` app, cf. rullzer/previewgenerator#135.

etc/ncp-config.d/nc-previews-auto.sh
nachoparker: Well, I think it is better to keep things intuitive, I try to avoid options if possible but that's me

nachoparker:
Just tell me when you are ready for me to apply this, and I will give it a test run as well

TomTurnschuh:
I didn't have time to continue the last days, but I will do soon.

nachoparker:
sure, no hurries at all. I don't have that much time myself

nachoparker:
sure, no hurries at all. I don't have that much time myself

Iolaum:
It may be that preview generator does not delete previews for deleted pictures, see:
https://github.com/rullzer/previewgenerator/issues/129

Just another thing to take into consideration.

nachoparker:
feedback issue https://github.com/nextcloud/nextcloudpi/issues/470

Iolaum:
In line with my previous comment that preview generator does not delete previews for deleted (and moved?) pictures I checked usage on the drive of my ncp instance. Previews take 158G out of the total 412G used on my hard drive :(

I like the previews, but I am seriously considering removing the app.

nachoparker:
@Iolaum I noticed that... it needs to be improved. I wish I had more time to help.

Iolaum:
Actually I went ahead and uninstalled the app. Even though the main developer is from the NextCloud team it doesn't look as this app is getting enough attention.

I agree with the time constraints. I removed the app because I decided that the time needed to babysit it, worry about it was not worth the benefit it provided. I actually like how little I needed to administer ncp after the upgrade to php7.2.

nachoparker:
I added a simple version of this feature here

https://github.com/nextcloud/nextcloudpi/commit/06073ed6d6d8e9d4a389ceff922e36a683e0e4f0

, let's keep improving that.

This https://github.com/rullzer/previewgenerator/issues/134 would also be nice to have in the future

PR #680, opened 2018-10-12
Update DDNS_spDYN.sh

TomTurnschuh:
- Switched from curl to wget
- Switched to spDYN for IP check
- Removed unused code

### DIFF
--- a/etc/ncp-config.d/DDNS_spDYN.sh
+++ b/etc/ncp-config.d/DDNS_spDYN.sh
@@ -3,7 +3,7 @@
 # spDYN setup for NextCloudPi
 #
 #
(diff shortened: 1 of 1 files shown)


hurrikan77:
Is there an possibility to test the new code now?

nachoparker:
@hurrikan77 I applied the PR to the `devel` branch (testing). It would be great if you applied it doing `sudo ncp-update devel` and play around with it a bit before I bring it to master.

@TomTurnschuh thanks!

hurrikan77:
@nachoparker @TomTurnschuh  
Thanks to both of you. My Cloud is back in the web. ;-)
The update process of the ip workes as expected.

PR #682, opened 2018-10-17
check app compatibility before before executing NC update

theCalcaholic:
This PR implements the following:

1) check supported versions of all enabled NC apps for compatibility with the new version before an update
2) if incompatible apps are found, write their names and version requirements to the logs and abort the update

There is one potential issue with the update check: It's executed pre-update, that means, if an app does not support the next NC version, but it's update (which would support the next version) does not support the current NC version, there is no way for the updater to succeed.
Example: Assuming the currently supported version of the news app supports NC 12-14 and NC 15 is going to be installed. There is an update to the news app, which adds support for NC 15 but does not support any previous version. This will cause the update to not be able to resolve (not even after manual app updates).

However I don't think that this case is very likely to occur. It could be avoided by doing the app compatibility check *after* the NC update has been executed and restoring the backup if any apps are not compatible - however I think this would be an additional source of errors, that isn't really worth it.

### DIFF
--- a/bin/ncp-update-nc
+++ b/bin/ncp-update-nc
@@ -65,40 +65,54 @@ echo "Available Nextcloud version $VER"
 # app compatibility check
 ##########################
 
-incompatible_apps=""
-for app_info in /var/www/nextcloud/apps/*/appinfo/info.xml
-do
-  version_xml=$(grep "$app_info" -e '< *nextcloud *min-version="[0-9][0-9]*" *max-version="[0-9][0-9]*" */>' | tr -d "\n")
-  for word in $version_xml
-  do
-    #echo "$word"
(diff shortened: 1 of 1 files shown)


theCalcaholic:
Do you require any action from my side for this PR? :)

nachoparker:
no, sorry just need some time to test it. NCP takes a lot of my time!

theCalcaholic:
I know :)
Take the time you need. :+1:

nachoparker:
@theCalcaholic I had to revert this. It is well done, but it only looks at the compatible versions of the currently installed apps.

It should see if there are updates available that are compatible with the new NC version instead. When I tried to use it it complained about almost every app.

screenshot_2019-01-08 nextcloudpi panel 2


BTW, `jq` would be perfect for this job :)

Issue #686, opened 2018-10-21
`nc-rsync` and `nc-rsync-auto` have different targets with the same configuration


b-abel:
`nc-rsync-auto` and `nc-rsync` have a small but significant difference. While `nc-rsync-auto.sh` syncs the 'datadirectory' of Nextcloud, the script `nc-rsync.sh` syncs only it's *content*.

I thought it would be a good idea to run `nc-rsync` manually and later - if the result looks good - take the same destination for `nc-rsync-auto`. But this lead to having the backup two times - one inside the other (poor little Pi had to transfer the 60GB a second time).

**Which behavior should be used for both commands?**

Reasoning:
* Syncing only the content - like `nc-rsync` does - has the advantage, that you can rename your 'datadirectory' without affecting the backup.
* On the other hand, keeping the behavior of `nc-rsync-auto` won't change anything on systems where this job is configured and runs regularly.

My take would be to go with the behavior of `nc-rsync-auto` because backwards compatibility is more important for regularly run jobs and because an already big 'datadirectory' probably won't  be renamed very often.

I would be happy to provide a PR.

nachoparker:
That's a good point. I applied your PR, thank you very much

PR #687, opened 2018-10-21
nc-rsync.sh: sync datadir, not only content (#686)

b-abel:
Changed rsync command to behave like nc-rsync-auto - as proposed in issue #686.


nachoparker:
Thanks for testing this! Applied

PR #688, opened 2018-10-21
ncp-web: fixed hover text for ncp wizard icon

b-abel:
hover text for ncp admin header icon of "NextCloudPi Wizard" was broken.


nachoparker:
good catch, thanks!

PR #690, opened 2018-10-23
Update Chinese translate.

chiyi4488:
This revision is mainly
1. Fix translations that cannot be displayed.
2. Add a translation.
3. Correct typos and grammar.
**Some translations are still in English, usually because I don't know how to use the feature, sorry.
**You may see that I modified the update file (ncp-update), this is just to let me test my file, and it has been corrected when the pull.
**I hope that this pull can be reviewed as soon as possible. This is just a translation. It should be difficult to make mistakes.
**It is recommended to directly pass the review, do not copy the file separately, because this pull is my server link my fork, test production, there is a certain guarantee that it can work normally, if it is copied, it may be a problem (I have This is the case, github seems to have add a symbol, which can not be displayed)

Thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ raspbian_boot
 ncp-web/wizard.cfg
 ncp-web/ncp-web.cfg
 docker-armhf/qemu-arm-static
+.vagrant/
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,50 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+#
+# Vagrantfile test the NCP curl installer
+#
+# Instructions: vagrant up; vagrant ssh
+#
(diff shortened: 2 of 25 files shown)


nachoparker:
I added it to `devel`, you can try it out `sudo ncp-update devel`

thanks!

chiyi4488:
Why I modified the file, after the pull, but can not display (fix the file can not be displayed)

too weird! Haha! Have to be revised again😄

chiyi4488:
Wait, nacho, can you please re-join my translation, which is a bit strange, my chiyi4488/nextcloudpi-1 is not the same as your nextcloud/nextcloudpi file. (Not modified, file when pull)

The picture below is just an easy-to-understand example. I am not sure why you can't display it, but I can actually display it, just confirmed by my server. Of course not only this file.

Issue #707, opened 2018-11-04
Better Clarity in NCP Admin settings related to database and datadir


wayneoutthere:
**Description**  
After losing many hours on two separate occasions I realized that I can file a bug here and maybe help improve clarity to save other people time.

**What we see now in the admin settings for data dir option** 
![ncp_datadir_admin_settings_confusion](https://user-images.githubusercontent.com/16578346/47965793-452d0400-e000-11e8-89c9-3cb6e9791f92.png)

**What we see now for the admin settings for database option**
![ncp_db_admin_settings_confusion](https://user-images.githubusercontent.com/16578346/47965796-6130a580-e000-11e8-95d1-633610a3b866.png)

## The problem
When an inexperienced user like me first goes into the these settings, it appears as though default lines in the fields are representing the **current** location of the db and datadir, but, it's not true.  After I ran 'sudo ncp-report' I discovered that's not where my db or datadir was pointing.

After wasting much time I realized that the intention of these two pages is if you plan to **change from the defaults** you enter new information here.

I surveyed two other people and they agreed this is not clear.

#### Option 1 - Show the current actual location of the two in the field
You could just put the path where they are now with the real names and locations and that would be more clear.  Then say 'if you wish to change the location, etc, just enter in the field and click 'run'"

#### Option 2 - give a long clear description in the page
Something like this:

>The current path in the field is **just a template**. 
>if you leave it blank and do not click 'run' your paths will remain as they were during your NCP >installation.  If you wish to change the location of the {datadir} just type in the new path.  Note >'USBdevice' is the NCP default label, blah blah blah"

I hope this is clear and will help the next version! :)

nachoparker:
implemented by @ovpc 

thanks both!

wayneoutthere:
coooool thx

PR #717, opened 2018-11-18
Clarify usage under info

ovpc:
Following request from user wayneoutthere https://github.com/nextcloud/nextcloudpi/issues/707
@nachoparker let me know if this is the correct way to add to the code: meaning doing pull request against devel?


nachoparker:
thanks!

nachoparker:
@ovpc yeah, that's perfect ;)

PR #736, opened 2018-12-06
Added PHPVER=7.2

ovpc:
To fix PATH in /etc/php/${PHPVER}/


ovpc:
See user input [here](https://help.nextcloud.com/t/need-help-with-nextcloudpi-installation-details-inside/42681/23)
Maybe PHPVER=7.2 needs to further up right at the start of nc-init? It did not seem to resolve the issue.

nachoparker:
that should work. Still I normally place it at the very top for visibility. I already pushed it in `v0.67.6`

thanks!

PR #749, opened 2018-12-21
Check for unit name of log2ram service before enabling/disabling it.

theCalcaholic:
This restores compatibility of the ncp app nc-ramlogs after the log2ram service was renamed on armbian.


nachoparker:
looks great, thank you!

PR #758, opened 2018-12-30
Update the systems data-fingerprint after restoration

TomTurnschuh:
Admin manual:

> After restoring a backup of your data directory or the database, you should always call `maintenance:data-fingerprint` once. This changes the ETag for all files in the communication with sync clients, allowing them to realize a file was modified.


nachoparker:
The file has been renamed, so I just pasted the two lines in the new place

https://github.com/nextcloud/nextcloudpi/commit/db8ded640ab3c6e0e3a86bfe874c9d67f65846c5

Thanks!

PR #771, opened 2019-01-13
Update nc-datadir.sh - fail2ban and tmp issues

PascalHeynmoeller:
Update fail2ban logpath for jail.local too and create  symboliclink from new data to old data


nachoparker:
Fixed it myself, thanks so much!

Issue #773, opened 2019-01-14
Wrong Regular Expression in `/run-parts.sh`


necopinus:
The regular expression on line 41 of `/run-parts.sh` (`docker-common/debian-ncp/run-parts.sh` in this repo) appears to be incorrect: `grep 000*` will match any string contain either `000` _or_ `00`. This means that when `dnsmasq` is enabled, for example, the startup service gets run twice, once after `000ncp` (which seems to be incorrect), and once with the other services (which appears to be the intention). This problem _should_ also effect the `noip` service, though I haven't tested that (I ran into this issue when working to get `dnsmasq` enabled within the container, see #772).

**Incorrect (Current) Method of Listing 000 Services (Line 41):**  
```
ls -1v /etc/services-enabled.d | grep 000*
```

**One Possible Correction for Listing 000 Services:**  
```
ls -1v /etc/services-enabled.d | grep ^000.*
```
The `^` ensures that we're only matching service files that _start_ with `000`, and using `.*` instead of `*` means that the character we're matching 0+ times is "any character", rather than "0".

**Suggested Further Improvement:**  
I might also change the call to find non-000 service files on line 46 to:
```
ls -1v /etc/services-enabled.d | grep -v ^000.*
```
The `-v` inverts the match, so this will match all of the service files that _weren't_ matched on line 41, which looks to be what you want to do. There's also a nice symmetry to these statements.

necopinus:
Pull request created, as requested.

Hoping I didn't mess anything up, as this is the first one of these that I've done.

nachoparker:
Thanks!

PR #781, opened 2019-01-21
Proposed fix for issue #773

necopinus:
Proposed fix for issue #773. To recap:

> The regular expression on line 41 of `/run-parts.sh` appears to be incorrect: `grep 000*` will match any string contain either 000 _or_ 00. This means that when dnsmasq is enabled, for example, the startup service gets run twice, once after `000ncp` (which seems to be incorrect), and once with the other services (which appears to be the intention).

The changes here:

1. Fix this regular expression so that _only_ services starting with 000 are run during the setup phase, and

2. Tweak the process for identifying all other enabled services to parallel this first change.

I have been running this change "in production" on my personal NextCloudPi server for a bit over a week now, with no observable issues or errors.

(FWIW, this is the first pull request I've ever done. Please let me know if there's something else that I need to do here.)

### DIFF
--- a/docker-common/debian-ncp/run-parts.sh
+++ b/docker-common/debian-ncp/run-parts.sh
@@ -38,12 +38,12 @@ EOF
 chmod +x /usr/local/sbin/update-rc.d
 
 # Iterate only over 000* entries which might setup environment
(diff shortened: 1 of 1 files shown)


nachoparker:
Looks good, thank you!

PR #785, opened 2019-01-23
Proposed fix for #784

### LINKED ISSUE #784: Update to v1.4.5 via ncp-update fails when cron.hourly is empty
When `/etc/cron.hourly` is empty (which can happen in docker containers), the update to v1.4.5 fails (because `chmod 755 /etc/cron.daily/* /etc/cron.hourly/*` on line 276 of `update.sh` takes an error). To account for this case, `update.sh` should test if various `/etc/cron.*` contain any files of interest before applying the fix.

Pull request incoming momentarily.

necopinus:
Test if `/etc/cron.*` directories have files before running `chmod` on them to prevent upgrade errors. Should fix #784.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -272,8 +272,9 @@ ncc config:system:set trusted_domains 1 --value=$ip
 EOF
 
   # fix Armbian cron bug
-  chmod 644 /etc/cron.d/*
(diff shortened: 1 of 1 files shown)


nachoparker:
ups, thank you!

PR #797, opened 2019-02-09
Update nc-static-IP.cfg

chiyi4488:
Just think it's better.


nachoparker:
Yeah, agree!

Thanks!

PR #802, opened 2019-02-12
add readme chinese translate

chiyi4488:


chiyi4488:
Zh file Should be mentioned last time, don't pull them, sorry

PR #835, opened 2019-03-11
 Inform the user about the current step

Tanarri:
In case of restore a big nextcloud instance the current infos, showed the user is missleading.
Think of the steps:
1) echo "restore datadir to $DATADIR..." (info to the user)
2)  echo "backing up existing $DATADIR to $DATADIR-$( date "+%m-%d-%y" )..." (--> Info for the user: a backup is performed. Because it only renames a directory the move will last only 1 second)
3) ..no more infos for the user.... the user think: Well, the step 2 lasts a llooonnngggg time.... But the program is doing the real restore of the $DATADIR

--> moved the "echo" a few lines down...

TanarRi


nachoparker:
thanks again :)

PR #840, opened 2019-03-13
Small typo

Tanarri:
'raspberry' is the password, and I think it should be quoted for better reading - like 'pi' a few words before in the same line.


nachoparker:
good catch

Issue #838, opened 2019-03-12
nc-restore.sh: Check free space before untar the backupfile.tar


Tanarri:
**Description**  
When a restore is done with nc-restore.sh the user needs at least as much free disc space as the backup-tar-file is large.

**Example**  
I had a backup-file "/media/USB-STICK/ncp-backup/nextcloud-bkp_xxxxxxxx.tar " on a usb-stick, which is 25 GB large.
In line 44 in https://github.com/nextcloud/nextcloudpi/blob/master/bin/ncp/BACKUPS/nc-restore.sh this .tar file is extracted in a subfolder, without checking, if the is enough free space on the drive.

It should be checked (or the user warned) that there has to be enough free disc space for the whole nc-data + nc-db

TanarRi

nachoparker:
Good idea, we do check during the backup 

https://github.com/nextcloud/nextcloudpi/blob/master/bin/ncp/BACKUPS/nc-backup.sh#L50

Feeling brave to port that code chunk to `nc-restore`?

Tanarri:
I have tried it yesterday, but have a small issue...
If the patch is ready I will make a pull-request

TanarRi

nachoparker:
cool, please let me know once it's ready

PR #842, opened 2019-03-14
 Check for free space in $TMPDIR before extracting tar file when doing restore

Tanarri:


Tanarri:
As dicussed in #838

nachoparker:
thanks!

PR #860, opened 2019-03-26
Info text corrected

Tanarri:
The info text says, that a snapshot will be "taken". That is wrong!
We are in restore-mode --> a snapshot will be "restored"


nachoparker:
lol true

Issue #844, opened 2019-03-15
nc-update-nextcloud path for temporary backup


paschaef:
In the update process of the nc-update-nextcloud functionality a temporary backup is made. However, this backup is always made at `/var/www`. 
Feature request: Make the location of the backup directory configurable in the ncp-web panel like in e.g. nc-backup. 

If nextcloudpi is hosting a lot of of data and is hosted on a raspberry pi there is not that much disk space (SD card) left for all the data to backup on the SD card. Data is located on an external drive. So it would be very helpful to make the temporary backup of nc-update-nextcloud also on this external drive.

nachoparker:
Hi,

The backup performed by `nc-update-nextcloud` is dataless, so it shouldn't be too big.

paschaef:
I am running a nextcloudpi docker instance and during the mentioned update process backup is failing due to disk size check... So updating with ncp-webpanel is not possible. 
Therefor this feature would be very appreciated.

Issue #858, opened 2019-03-26
Docker: nc-backup: free space check failed


langfingaz:
**System information**
Arch Linux with docker and docker-compose installed. Formatted with btrfs.
Running NextcloudPi form Docker_x86.

**Problem Found**
When I run nc-backup (excluding data, without compression, path=/data/ncp-backups) I get this eror:
```
[ nc-backup ]
check free space...
free space check failed. Need 1731093668 Bytes
Maintenance mode already disabled
```

Even if there's enough space left as can bee seen by `df -h` from inside the nextcloudpi container.
```
root@57159478c0ec:/# df -h
Filesystem         Size  Used Avail Use% Mounted on
/dev/mapper/crypt  238G   22G  216G   9% /
tmpfs               64M     0   64M   0% /dev
tmpfs              3.9G     0  3.9G   0% /sys/fs/cgroup
/dev/mapper/crypt  238G   22G  216G   9% /data
shm                 64M     0   64M   0% /dev/shm
tmpfs              3.9G     0  3.9G   0% /proc/asound
tmpfs              3.9G     0  3.9G   0% /proc/acpi
tmpfs              3.9G     0  3.9G   0% /proc/scsi
tmpfs              3.9G     0  3.9G   0% /sys/firmware
/dev/mapper/data   2.8T  1.7T  1.2T  60% /data/nextcloud/data
```

**Research**
The `ncp-backup` script checks the space left with `free=$( df "$destdir" | tail -1 | awk '{ print $4 }' )
` . Running it with my chosen backup path returns only approx. 200mb of space instead of approx. 216G:
```
root@57159478c0ec:/# df "/data/ncp-backups" | tail -1 | awk '{ print $4 }'  
226067952
```
Looks like the above method does not work properly in this use case (inside docker container on btrfs).
Any ideas how you could still perform the space-left check?

nachoparker:
what does this give you `df "/data/ncp-backups"`?

langfingaz:
```
root@6f2cf01ef85b:/data# df "/data/ncp-backups"
Filesystem        1K-blocks     Used Available Use% Mounted on
/dev/mapper/crypt 249492804 23350888 224859800  10% /data
```
So would guess it's off by the factor 1000 in my case.

nachoparker:
Yes, that's wrong. Please run `sudo ncp-update devel` and verify that it is fixed

langfingaz:
Thanks for the quick reply!

[...]

PR #864, opened 2019-03-29
fix: corrected diskspace calculation in case of docker usage

paschaef:
As mentioned in  #844 during backup in the ncp-update process there is the possiblity that no diskspace is available for the backup. Also it was mentioned that the backup is dataless and does not require a lot of space. 
However, running in a docker configuration the data-directory is a subdirectory of the `basedir` and hence size of `datadir`gets counted too eventhough `includedata` is disabled. 
This pull request aimed to correct the calculation for the free/occupied diskspace in docker configuration. If not in docker configuration calculation is done as before.

### DIFF
--- a/bin/ncp/BACKUPS/nc-backup.sh
+++ b/bin/ncp/BACKUPS/nc-backup.sh
@@ -38,25 +38,16 @@ fail()   { local ret=$?; echo "Abort..."  ; rm -f "${dbbackup}" "${destfile}"; $
 trap cleanup EXIT
 trap fail INT TERM HUP ERR
 
-echo "check free space..." # allow at least ~100 extra MiB
+echo "check free space..." # allow at least ~500 extra MiB
 mkdir -p "$destdir"
 dsize=$(du -s "$datadir" | awk '{ print $1 }')
 nsize=$(du -s "$basedir/nextcloud" | awk '{ print $1 }')
-if [[ -f /.docker-image ]] 
+margin=$((500*1024))  # safety margin to allow additional space
+if [[ "$includedata" == "yes" ]]
 then
-  if [[ "$includedata" == "yes" ]]
-  then
-    size=$((nsize + 100*1024))
-  else #datadir is inside $basedir/nextcloud therefor substract
-    size=$((nsize - dsize + 100*1024))
-  fi
-else
-  if [[ "$includedata" == "yes" ]]
-  then
-    size=$((nsize + dsize + 100*1024))    
-  else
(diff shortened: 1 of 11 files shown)


nachoparker:
good catch

nachoparker:
related to conversation here https://github.com/nextcloud/nextcloudpi/issues/858#issuecomment-478335791

nachoparker:
This should not be tied to `/.docker-image`. If we are not using `INCLUDEDATA`, then we should exclude the size of the datadir from the calculation.

Also we should add some more space for the database dump, as per the conversation above.

Thanks!

paschaef:
So for clarification of the directory structre (because I am only using the dockerized version). The directory structure looks as follows for **both** (docker, non-docker) cases:
```
/var/www/nextcloud/
/var/www/nextcloud/$(datadir)
```
Am I right?

Then the proposal is to remove the outer most `if` condition with `/.docker-image` and to do only this part: 
```bash
  if [[ "$includedata" == "yes" ]]
  then
    size=$((nsize + 100*1024))
  else #datadir is inside $basedir/nextcloud therefor substract
    size=$((nsize - dsize + 100*1024))
  fi
```

Any suggestions on how to estiamte the size of the database dump? Or would you like to make a (larger) hardcoded estimate as it is now?

nachoparker:
hi

Yes, the dirs are `"$BASEDIR/nextcloud` and `"$BASEDIR"/nextcloud/data` in the docker or regular case.

Your proposed solution looks good to me. 

About the db dump... it's hard to tell. My dbdir is 1GiB and my dbdump is 300MiB, so I think that the easiest is to go from 100 to 500MiB extra

paschaef:
Here we are.

nachoparker:
Looks good! We only need to tweak the calculations like in this commit https://github.com/nextcloud/nextcloudpi/commit/c18273a0f69669dc8846e786ded7501cd46aeb96, which was fixed after you started working on this.

paschaef:
This is done! Hope it is correct in this version now.

nachoparker:
looks perfect now :)

paschaef:
great :)

nachoparker:
thanks!

hunhejj:
I still have the above issue.

Nextcloudpi version: `1.31.0.`, running in docker.

Trying to upgrade nextcloud from `18.0.7.1` to `19.0.4`.

Fails with the same error as mentioned in #858: `free space check failed.  Need 873366317 Bytes`. Checked and there is way more space than that.

PR #865, opened 2019-03-29
Adjust path for restore

Tanarri:
The suggested path to the backupfile for restore should be the same as the path used in backup.

Actual:
Backuppath in nc-backup-auto.cfg & nc-backup.cfg: /media/USBdrive/**ncp-backups**
Backuppath in nc-restore.cfg: /media/USBdrive/

This path adjust the paths in .cfg

TanarRi


nachoparker:
merged locally, thanks!

PR #883, opened 2019-04-14
Updated spDYN.sh

TimmThaler:
- As suggested by nachoparker in https://help.nextcloud.com/t/spdyn-updater-not-working/32199/26 over a year ago, I finally adjusted the setup script to put the bash file into `/usr/local/bin/` instead of `/usr/local/etc`.
- Also I adjusted the "Usage" section a bit to fit ncp usage with IPv6

I think, that should be better :)

### DIFF
--- a/bin/ncp/NETWORKING/spDYN.sh
+++ b/bin/ncp/NETWORKING/spDYN.sh
@@ -10,8 +10,7 @@
 #
 
 
-INSTALLDIR=spdnsupdater
-INSTALLPATH=/usr/local/etc/$INSTALLDIR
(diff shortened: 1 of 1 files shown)


nachoparker:
thanks! we also need to move the file from the old to the new location in `update.sh` or it will fail for people that already use it.

nachoparker:
superseeded by https://github.com/nextcloud/nextcloudpi/pull/929

TimmThaler:
Sorry, that I didn't reply. I had too much to do over the last months to do it properly myself ... :( 

Thank you for your work!

PR #889, opened 2019-04-20
Rename configuration variables into self-documenting ones

### LINKED ISSUE #886: Rename configuration variables into self-documenting ones
Especially when it comes to number-based configuration variables, I often had to look into the source code to be 100% sure what they meant. That is why I suggest to rename some of them to make their purpose more clear to other users. 

- Backups
  - nc-backups-auto
    - Backup days -> Backup periodicity (in days)
    - Backup limit -> Number of backups to keep
  - nc-backup
    - Backup limit -> Number of backups to keep
  - nc-rsync-auto
    - Sync days -> Sync periodicity (in days)
  - nc-snapshot-sync
    - Sync days -> Sync periodicity (in days)
- Config
  - nc-previews-auto
    - Run time -> Runtime (in minutes)
  - nc-scan-auto
    - Scan interval -> Scan periodicity (in minutes)
- Networking
  - freeDNS
    - Update Interval -> Update periodicity (in minutes)
  - no-ip
    - Time -> ?

I used the word periodicity a lot of times, maybe there is a better one, but it would be great if it would be the same for all entries which occur periodically :)

Hofer-Julian:
Fixes #886

### DIFF
--- a/etc/ncp-config.d/freeDNS.cfg
+++ b/etc/ncp-config.d/freeDNS.cfg
@@ -26,7 +26,7 @@
     },
     {
       "id": "UPDATEINTERVAL",
-      "name": "Update Interval",
+      "name": "Update periodicity (in minutes)",
       "value": "30",
       "suggest": "30"
     }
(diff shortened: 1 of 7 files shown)


nachoparker:
thanks!

PR #898, opened 2019-05-06
Dockerfiles for ARM 64 bit

VoidStructure:
Adds the necessary files for ARM 64 bit Docker images.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ raspbian_boot
 ncp-web/wizard.cfg
 ncp-web/ncp-web.cfg
 docker-armhf/qemu-arm-static
+docker-aarch64/qemu-aarch64-static
 .vagrant/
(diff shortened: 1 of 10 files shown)


nachoparker:
Thanks! this is great!

Actually this is the perfect moment to unify all architectures in one single Dockerfile. I have been wanting to do this for simplicity for a while but never got to spend some time on it.

Let me do some changes first before we merge this.

I am sorry that you already did all this, next time it's better to just talk about it so we can make sure we are on the same page :)

nachoparker:
Hi there

I have refactored the docker builds. There was too much duplication.

https://github.com/nextcloud/nextcloudpi/commit/103b4811d16cadad15806583308bddfe1401a293

Could you try building an `arm64` container and seeing if all is ok?

In order to build now it would be going to the `devel` branch and doing

```
./build-docker.sh arm64
```

VoidStructure:
Hey, 
I built the docker images and tested them on a rock64, it all worked flawlessly. 
We can probably remove this pull request then.

nachoparker:
perfect! thanks very much

sorry that we are going to close this, but I wanted to restructure the code to avoid a third copy of the install code. Now all three docker images share the same Dockerfiles

PR #907, opened 2019-05-18
Some Pt Translate

juaryR:
Some translations to   Portuguese.


nachoparker:
great thanks! will test soon

PR #928, opened 2019-06-21
Update ncp-update to fix backward update mechanism

FaniD:
Backward update mechanism fixed, using binary
search to find the starting update file needed and
then execute the following update files.
Update files should be located inside updates
folder under the name update_x_y_z_.sh
where x_y_z represents the version of ncp
which the current version will be updated to.
A few sample files are generated inside the updates
folder.


bin/ncp-update
```diff
@@ -20,7 +20,159 @@
   [[ -f /.ncp-image ]] || cd "$TMPDIR"         # update locally during build
 
   echo -e "Performing updates"
-  ./update.sh || exit 1
+
```
nachoparker: let's leave `ncp-update` the way it was and move all this code to `update.sh`

bin/ncp-update
```diff
@@ -20,7 +20,159 @@
...
+  done < <( ls -1 updates | sort -V)
+
+  starting_checkpoint=0
+  len=${#updates_list[@]}
+  end_of_list=$(expr $len - 1)
+
+  # The latest checkpoint is the newer version in updates dir
+  latest_checkpoint=${updates_list[$end_of_list]}
+
+  # Compare current version with latest checkpoint to see if we need backwards updates
```
nachoparker: let's make a function in `etc/library.sh` to that returns true if version A is more recent than version B

updates/update_1_10_11_.sh
```diff
@@ -0,0 +1,263 @@
+#!/bin/bash
```
nachoparker: These files should only contain the part after `## BACKWARD FIXES ( for older images )`.

The updates folder should only contain one file initially with the contents of the current `update.sh` after `## BACKWARD FIXES ( for older images )`.

bin/ncp-update
FaniD: I wasn't sure about the correct position inside update.sh, so I created a distinct script of the backward mechanism so that it can be moved easier to the right place.

updates/update_1_10_11_.sh
FaniD: I created one script - the contents of *BACKWARD FIXES* part and named it 1.10.11.sh (the version before latest cleanup of update.sh). Feel free to fill the updates folder, just keep in mind that my mechanism, as it is written right now, works with files of the following format: x.y.z.sh where x.y.z is the version that the system will update to.

update.sh
```diff
@@ -49,6 +49,8 @@ cp etc/library.sh /usr/local/etc/
 
 source /usr/local/etc/library.sh
 
+./backward_updates.sh
```
nachoparker:
```suggestion
./run_update_history.sh "updates"
```

What do you think about that name? also it would be nice to have the folder be a parameter

etc/library.sh
```diff
@@ -230,6 +230,33 @@ function persistent_cfg()
   ln -s "$DST" "$SRC"
 }
 
+function is_more_recent_than()
+{
+  local version_A="$1"
+  local version_B="$2"
+ 
+  MAJOR_A=$( echo ${version_A} | cut -d. -f1 )
```
nachoparker:
```suggestion
  local major_a=$( echo ${version_A} | cut -d. -f1 )
```

(there's more in this function)

backward_updates.sh
```diff
@@ -0,0 +1,111 @@
...
+latest_checkpoint=${updates_list[$end_of_list]}
+latest_checkpoint_version=$( basename ${latest_checkpoint} .sh )
+current_version=$( grep -oP "\d+\.\d+\.\d+" /usr/local/etc/ncp-version | cut -d'v' -f1 )
+
+# Compare current version with latest checkpoint
+# If the current version is more recent than the latest checkpoint there is no need for backward updates
+
+BACKWARD_UPDATES=$( is_more_recent_than ${latest_checkpoint_version} ${current_version} )
+
+if $BACKWARD_UPDATES ; then
```
nachoparker:
```suggestion
if  is_more_recent_than ${latest_checkpoint_version} ${current_version}; then
```

backward_updates.sh
```diff
@@ -0,0 +1,111 @@
...
+set -e
+
+source /usr/local/etc/library.sh
+
+# Get the array of updates dir
+# The files in updates dir are sorted by tag number
+# Files names follow the syntax of a tag: x.y.z.sh
+i=0
+while read line ; do
+  updates_list[ $i ]="$line"
```
nachoparker:
```suggestion
  updates_list +=("$line")
```
you don't need the `i` variable

backward_updates.sh
```diff
@@ -0,0 +1,111 @@
...
+  (( i++ ))
+done < <( ls -1 updates | sort -V)
+
+starting_checkpoint=0
+len=${#updates_list[@]}
+end_of_list=$(expr $len - 1)
```
nachoparker:
```suggestion
end_of_list=$((len - 1))
```

backward_updates.sh
```diff
@@ -0,0 +1,111 @@
...
+  while [ $lower_bound -le $upper_bound ]; do
+    x=$(expr $upper_bound + $lower_bound)
+    mid=$(expr $x / 2 )
+
+    #Compare mid's version with current version
+
+    mid_version=$( basename ${updates_list[$mid]} .sh )
+    apply_update=$( is_more_recent_than ${mid_version} ${current_version} )
+
+    if $apply_update ; then
```
nachoparker:
```suggestion
    if is_more_recent_than ${mid_version} ${current_version}; then 
```

update.sh
FaniD: Sure! Sounds good!

run_update_history.sh
```diff
@@ -5,19 +5,18 @@
 set -e
 
 source /usr/local/etc/library.sh
+updates="$1"
```
nachoparker: let's check for the existence of this folder

run_update_history.sh
```diff
@@ -0,0 +1,103 @@
...
+	break
+      fi
+    fi
+  done
+
+  # Starting checkpoint has been found so update the system for the rest updates
+
+  for(( i=${starting_checkpoint}; i<=${end_of_list}; i++)); do
+    update_file=${updates_list[i]}
+    tag_update=$( basename ${update_file} .sh )
```
nachoparker:
```suggestion
    tag_update=$( basename "${update_file}" .sh )
```

run_update_history.sh
```diff
@@ -0,0 +1,103 @@
...
+    echo "Update system to version ${tag_update}"
+    ./${updates}/${update_file} || exit 1
```
nachoparker:
```suggestion
    ./${updates}/"${update_file}" || exit 1
```

run_update_history.sh
```diff
@@ -0,0 +1,103 @@
...
+  done
```
nachoparker: we should update the system version after each update, so that if update number 3 fails, we don't have to run updates 1 and 2 again, which have already been successful.

run_update_history.sh
FaniD: I did changes like that to the whole script to get closer to your programming style

run_update_history.sh
FaniD: Smart! Done :)

etc/library.sh
```diff
@@ -230,6 +230,33 @@ function persistent_cfg()
   ln -s "$DST" "$SRC"
 }
 
+function is_more_recent_than()
+{
+  local version_A="$1"
+  local version_B="$2"
+ 
+  local major_a=$( echo "$version_A" | cut -d. -f1 )
```
nachoparker:
```suggestion
  local major_a=$( cut -d. -f1 <<<"$version_A" )
```

etc/library.sh
```diff
@@ -230,6 +230,33 @@ function persistent_cfg()
...
+  local patch_a=$( echo "$version_A" | cut -d. -f3 )
+
+  local major_b=$( echo "$version_B" | cut -d. -f1 )
+  local minor_b=$( echo "$version_B" | cut -d. -f2 )
+  local patch_b=$( echo "$version_B" | cut -d. -f3 )
+
+  # Compare version A with version B
+  # Return true if A is more recent than B
+
+  if [ "$major_b" -gt "$major_a" ]; then
```
nachoparker:
```suggestion
  if [[ "$major_b" > "$major_a" ]]; then
```

run_update_history.sh
```diff
@@ -0,0 +1,108 @@
...
+
+# Backward updates mechanism
+
+set -e
+
+source /usr/local/etc/library.sh
+
+# Updates folder contains the "history" of updates
+updates="$1"
+[ -d "$updates" ] || exit 1
```
nachoparker:
```suggestion
[ -d "$updates" ] || { echo "$updates does not exist. Abort" >&2; exit 1; }
```

etc/library.sh
FaniD: I actually left this like this in order to be in accordance with ncp-test-updates.

FaniD:
I've tested the mechanism plenty of times, and it certainly succeeds in finding the right update to start with and execute the updates sequentially. It's now up to the updates script files to make a successful update :)

nachoparker:
Many thanks Fani. Thumbs up for merging then?

FaniD:
Sure! Glad to contribute to ncp and thanks for the coding tips I got from you! :)

nachoparker:
merged manually

nachoparker:
@FaniD I had to make a bunch of simple adjustments. Please see https://github.com/nextcloud/nextcloudpi/commit/88da901eddff125b3e71d9706fbe02edffc659b3

PR #929, opened 2019-06-23
removing unused IPV6 argument in spDYN.sh

theGreatWhiteShark:
Since #883 is still open and I run into the same problems with spdyn (which took me quite some time to figure out what was going wrong), maybe another take helps to resolve the bug (more quickly).

Right now the `IPv6` argument in the spdnsUpdater.sh script (created using spDYN.sh) is not covered by whiptail. This causes raspis having only a IPv6 address to be unable to updated their dynamic ip address provided by spdyn.

Instead of either querying `http://checkip4.spdyn.de` or `http://checkip6.spdyn.de` I would advocate to query just `http://checkip.spdyn.de`. It returns the IPv4 or v6 address if the client has just one of them and the Ipv6 address if the client has both.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,8 +2,8 @@
 
 Check this wiki article if you want to learn how you can help.
 
-[Ways to contribute](http://docs.nextcloudpi.com/en/latest/Maintain/Contribute/)
+[Ways to contribute](https://docs.nextcloudpi.com/en/contribute/)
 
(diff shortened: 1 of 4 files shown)


nachoparker:
looks good, thanks!

nachoparker:
cherry-picked

TimmThaler:
Sorry, that I didn't reply. I had too much to do over the last months to do it properly myself ... :( 

Thank you for your work!

PR #936, opened 2019-07-04
copy NCP app to nextcloud directory

qrest:
The update script only updates the NCP app at `/var/www/ncp-app` but not at `/var/www/nextcloud/apps/nextcloudpi`. This way it is not actually used by Nextcloud. This is fixed by this pull request.


nachoparker:
good catch. Thanks!

Issue #913, opened 2019-05-30
Problem with ufwban


musklor:
Hi, I can't use fail2ban anymore.

Ufw is activated
Modsecurity is activated
I've clean files and made a clean install but always the same issue.
Datadir : external hdd

Ncp-panel :

[ fail2ban ]
System config value loglevel set to string 2
System config value log_type set to string file
Job for fail2ban.service failed because the control process exited with error code.
See "systemctl status fail2ban.service" and "journalctl -xe" for details.
fail2ban enabled

Journal :

[...]

PR #937, opened 2019-07-05
make sure ufw.log always exists for fail2ban

dhiltonp:
Full discussion in #913 
Not fully tested as I'm currently remote.

### DIFF
--- a/bin/ncp-dist-upgrade
+++ b/bin/ncp-dist-upgrade
@@ -47,6 +47,7 @@ set -e
 
 # update sources
 sed -i 's/stretch/buster/g' /etc/apt/sources.list
+sed -i 's/stretch/buster/g' /etc/apt/sources.list.d/*
 rm -f /etc/apt/sources.list.d/php.list
 
 # install latest distro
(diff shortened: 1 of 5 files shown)


nachoparker:
Many thanks! You are right. This would be over-written by an update. I think that the correct place to introduce the `touch` command would be `/etc/systemd/system/fail2ban.service.d`, which is a system drop-in unit rule that I already use for fail2ban and does not get replaced during updates.

dhiltonp:
Those are conf files, the don't seem executable?

nachoparker:
They are unit files. You can try creating a file named `/etc/systemd/system/fail2ban.service.d/touch-ufw-logfile.conf`

with these contents

```
[Service]
ExecStartPre=/bin/touch /var/log/ufw.log
```

The logfile should always appear even if UFW is disabled.

dhiltonp:
I'll be able to test that change in about a week.

dhiltonp:
Updated. I've manually applied these changes and tested the result and it works fine.

There are still some things I don't quite understand:

UFW.sh has something to "disable logging to kernel":
grep -q maxsize /etc/logrotate.d/ufw || sed -i /weekly/amaxsize2M /etc/logrotate.d/ufw

Looking at things, maybe we should do `sed -i 's/#\& stop/\& stop /' /etc/rsyslog.d/20-ufw.conf` instead? It would log UFW messages to ufw.log only, removing UFW logs from /var/log/messages.

I also don't understand how `/var/log.hdd/` fits in with all of this...

nachoparker:
Thanks! that comment does not explain what that line does. We are just limiting the  size of the logfile.

`/var/log.hdd` is part of the ram2logs scheme

PR #939, opened 2019-07-06
do not enable nextcloud app in update script

qrest:
When creating new images `update.sh` gets called. At this point Nextcloud is not yet fully installed and thus `app:enable` fails.


qrest:
Seems like a made a mistake here. Just ignore this PR.

PR #940, opened 2019-07-06
do not enable nextcloud app in update script

qrest:
When creating new images `update.sh` gets called. At this point Nextcloud is not yet fully installed and `thus app:enable` fails. This PR just remove this one line.


nachoparker:
Ah good catch. It shouldn't need to be enabled again anyway.

Are you playing with the build system? that's great, we need a lot of help with that. Thanks

qrest:
Yeah, I noticed it in the process of upgrading to buster.

nachoparker:
I have a pi4 image with buster in testing but we haven't moved officially to buster yet.

Anyhow, with the recent changes it should just work in both buster and stretch

PR #946, opened 2019-07-11
Remove Expose Ports

langfingaz:
If you start an docker image via `docker` or `docker-compose` all ports which are exposed through inside the Dockerfile are always opened! The user cannot disable it.

See `sudo docker ps` after starting nextcloudpi-docker: Ports 80, 443 and 4443 are always exposed on the host.

This even bypasses the **UFW** setup on the host system as docker directly modifies the ip-tables. There exist some issues on the Docker repo discussing this problem. But the easiest "workaround" is not to force-expose any ports.

Launching the Docker image normally or via docker-compose should not be affected. If the user has for example 

```
    ports:
     - 80:80
     - 443:443
     - 4443:4443
```

included in his compose file, he won't notice this change.

What do you think of this? Is there any reason I missed for exposing the ports via Dockerfile?

Greetings,
Daniel

### DIFF
--- a/docker/nextcloudpi/Dockerfile
+++ b/docker/nextcloudpi/Dockerfile
@@ -68,6 +68,3 @@ echo "${ncp_ver}" > /usr/local/etc/ncp-version
 COPY docker/nextcloudpi/000ncp /etc/services-enabled.d/
 COPY bin/ncp/CONFIG/nc-init.sh /
 COPY etc/ncp-config.d/nc-init.cfg /usr/local/etc/ncp-config.d/
-
-# 4443 - ncp-web
-EXPOSE 80 443 4443


langfingaz:
I have to admit that there are two other ways how the user can stop docker from manipulating the ip-tables, but in my opinion this is the less natural way to solve this issue:

>  Fortunately, there's a way to fix this. Go back to the terminal on your Docker server and issue the command sudo nano /etc/default/docker and add the following line:
> `DOCKER_OPTS="--iptables=false"`

Reference:
* https://www.techrepublic.com/article/how-to-fix-the-docker-and-ufw-security-flaw/


> sudo nano /lib/systemd/system/docker.service
> `ExecStart=/usr/bin/dockerd -H fd:// --iptables=false`
>
> `sudo systemctl daemon-reload`
> `sudo systemctl restart docker`

Reference:
* https://github.com/moby/moby/issues/22054#issuecomment-217543551
* https://github.com/moby/moby/issues/22054#issuecomment-217702262

nachoparker:
I don't see a problem with removing the EXPOSE

langfingaz:
Ok great.
Sorry for not signing off the commit. I just don't see any option to do this now afterwards.

nachoparker:
Please send PRs against `devel` in the future. Thanks for the PR!

Issue #938, opened 2019-07-06
dhcpcd does not properly start on boot or reboot - must be started manually


Halloran:
Each time the machine is booted, dhcpcd fails to start properly, and therefore, the it does not get a proper ip v4 address. In order to resolve, I need to manually log in @ nextcloudpi.local, and then `sudo dhcpcd` to start dhcpcd.

This works fine, but obviously it prevents unattended reboots, which may happen if power is interrupted, etc.

**System information**

<--! Paste this in GitHub report -->

<details>
<summary>NextCloudPi diagnostics</summary>

```

NextCloudPi version  v1.14.0
NextCloudPi image    NextCloudPi_09-28-17
distribution         Raspbian GNU/Linux 9 \n \l
automount            yes
USB devices          sda 
datadir              /media/nextCloudDrive/ncdata
data in SD           no
data filesystem      ext2/ext3
data disk usage      769G/3.6T
rootfs usage         5.0G/15G
swapfile             /media/USBdrive/swap
dbdir                /media/USBdrive/ncdatabase
Nextcloud check      ok
Nextcloud version    15.0.8.1
HTTPD service        up
PHP service          up
MariaDB service      up
Redis service        up
Postfix service      up
internet check       ok
port check 80        open
port check 443       open
IP                   ***REMOVED SENSITIVE VALUE***
gateway              ***REMOVED SENSITIVE VALUE***
interface            eth0
certificates         ***REMOVED SENSITIVE VALUE***
NAT loopback         no
uptime               10min

```
</details>

<details>
<summary>Nextcloud configuration</summary>

```

{
    "system": {
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": {
            "0": "localhost",
            "5": "nextcloudpi.local",
            "1": "192.168.1.20",
            "3": "hallorancloud.ddns.net",
            "4": "hallorancloud.ddns.net",
            "7": "nextcloudpi",
            "8": "nextcloudpi.lan"
        },
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "overwrite.cli.url": "https:\/\/192.168.1.20\/",
        "dbtype": "mysql",
        "version": "15.0.8.1",
        "dbname": "***REMOVED SENSITIVE VALUE***",
        "dbhost": "***REMOVED SENSITIVE VALUE***",
        "dbport": "",
[...]

PR #952, opened 2019-07-20
prevent user query in ncp update

SoongJr:
web interface was unable to conclude the update because it was asking for user input.
See https://github.com/nextcloud/nextcloudpi/issues/938


nachoparker:
many thanks. Already fixed :)

Issue #965, opened 2019-08-07
Missing PHP module in new Nextcloud Pi docker image


paschaef:
**System information**
 Nextcloud Pi Docker image `NextCloudPi_docker_07-20-19`
-  On Raspberry Pi (nextcloudpi-armhf with image ID: `3b832c3ddae7`)
- On x86 machine (nextcloudpi-x86 with image ID: `547600d6f319` )

**Problem**
After a fresh install of nextcloud pi in the above mention docker containers, nextcloud reports a warning about a missing PHP module `imagick` in the administrators overview tab (see image). 
![php_missing](https://user-images.githubusercontent.com/45995338/62600780-05c57800-b8f0-11e9-86d8-1a9a709224b8.PNG)

After installation the installation inside the docker container with
``` bash
apt-get install php-imagick
```
and a restart of the nextcloud container  the warning disappears. However, this change is not persistant in the container. After a complete shutdown of the container the effect of the above command has gone.

This php module should be already included in the docker image, so that no warning is generated in nextcloud admin page.

nero120:
@paschaef see [this](https://github.com/nextcloud/server/issues/13099) discussion in the nextcloud server repo, not so much a nextcloudpi issue.

paschaef:
@nero120 agree. As mentioned in https://github.com/nextcloud/nextcloudpi/pull/966#issuecomment-519026102 I wasn't aware that this is already an ongoing nextcloud issue.

PR #966, opened 2019-08-07
lamp.sh: install additionally needed php-imagick module for nextcloud 16

paschaef:
I found the file where the missing php module has to be installed to fix issue #965.
Nextcloud runs without complaining about missing php-imagick module.

### DIFF
--- a/bin/ncp-dist-upgrade
+++ b/bin/ncp-dist-upgrade
@@ -47,6 +47,7 @@ set -e
 
 # update sources
 sed -i 's/stretch/buster/g' /etc/apt/sources.list
+sed -i 's/stretch/buster/g' /etc/apt/sources.list.d/*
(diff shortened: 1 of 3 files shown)


ovpc:
I thought it was intentionally removed by NC devs, some security issue with imagick. ?!

paschaef:
I did not know that imagick under discussion for security issues. Found that according to https://github.com/nextcloud/server/issues/13099#issue this is still an ongoing discussion.

As the origin of this PR is #965 where the recommendation nextcloud is to install imagick. However, this recommendation is target to the discussion in the above link. 

Either imagick should be installed or the recommendation for imagick from nextcloud should be removed...

nachoparker:
Thanks for the help. This went under my radar. This should be fixed upstream Nextcloud, so it is not a NCP issue.

We'll see how this imagick drama ends :)

PR #974, opened 2019-08-17
Check if apt/sources.list.d directory is empty 

qrest:
The directory ``/etc/apt/sources.list.d`` is empty on my docker installation. It prevented me from updating to ncp 1.15.3. With the check the update goes through successfully.


nachoparker:
thanks! I applied the fix

PR #990, opened 2019-09-04
added cp commands for folder 'ncp-app' and 'ncp-web'

steubner:
When installing with 'install.sh' script I had some issues due to missing folders 'ncp-app' & 'ncp-web' in '/var/www' folder. A copy command for both folders has been added.


nachoparker:
many thanks but those files are already copied in adifferent place.

What happened is that there was a point where installation failed and as a consequence those files were never copied and thus the error you saw

PR #991, opened 2019-09-05
added cp commands for folders 'ncp-app' and 'ncp-web'

steubner:
When installing with 'install.sh' script I had some issues due to missing folders 'ncp-app' & 'ncp-web' in '/var/www' folder. A copy command for both folders has been added.


nachoparker:
https://github.com/nextcloud/nextcloudpi/pull/990#issuecomment-528921162

Issue #959, opened 2019-07-25
Request Entity Too Large


musklor:
Hello, 
Nextcloudpi : v1.15.3
Error : Request Entity Too Large
When : upload
Board : RPI3B

Post_max_size = 10G
upload_max_filesize = 10G

!-- Paste the output of `sudo ncp-report` -->
<details>
<summary>NextCloudPi diagnostics</summary>

```

NextCloudPi version  v1.15.3
NextCloudPi image    NextCloudPi_03-09-19
distribution         Raspbian GNU/Linux 10 \n \l
automount            yes
USB devices          sda sdb 
datadir              /media/CloudDrive/ncdata
data in SD           no
data filesystem      btrfs
data disk usage      51G/466G
rootfs usage         2,2G/57G
swapfile             /var/swap
dbdir                /var/lib/mysql
Nextcloud check      ok
Nextcloud version    16.0.3.0
HTTPD service        up
PHP service          up
MariaDB service      up
Redis service        up
Postfix service      up
internet check       ok
port check 80        open
port check 443       open
IP                   ***REMOVED SENSITIVE VALUE***
gateway              ***REMOVED SENSITIVE VALUE***
interface            wlan0
certificates         ***REMOVED SENSITIVE VALUE***
NAT loopback         yes
uptime               2:08

```
</details>

<details>
<summary>Nextcloud configuration</summary>

```

{
    "system": {
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": {
            "0": "localhost",
            "11": "86.196.246.245",
            "1": "192.168.1.130",
            "5": "nextcloudpi.local",
            "7": "nextcloudpi",
            "8": "nextcloudpi.lan",
            "4": "amoi.my-firewall.org"
        },
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "dbtype": "mysql",
        "version": "16.0.3.0",
        "overwrite.cli.url": "https:\/\/amoi.my-firewall.org\/",
        "dbname": "***REMOVED SENSITIVE VALUE***",
        "dbhost": "***REMOVED SENSITIVE VALUE***",
        "dbport": "",
        "dbtableprefix": "oc_",
        "mysql.utf8mb4": true,
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "installed": true,
[...]

PR #993, opened 2019-09-08
Increased modsecurity bodynofileslimit so larger files can be synced

Iolaum:
Temporary(?) fix to #959

Needs testing. I added functionality in `update.sh` to update the config for people who already have modsecurity running, please check that I haven't overlooked something (or misunderstood chow `check_distro` works). The intent is that the fix would only apply to people who are in buster and have enabled, at some point, modsecurity.

People who enable modsecurity for the first time get the updated config file.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -186,59 +186,6 @@ EOF
 chmod +x /etc/update-motd.d/30ncp-dist-upgrade
 }
 
-# Update modsecurity config file only if user is already in buster and is used.
-# https://github.com/nextcloud/nextcloudpi/issues/959
-check_distro "$NCPCFG" && {
-  [[ -f /etc/modsecurity/modsecurity_crs_99_whitelist.conf ]] && {
-    cat > /etc/modsecurity/modsecurity_crs_99_whitelist.conf <<EOF
-<Directory $NCDIR>
-  # VIDEOS
-  SecRuleRemoveById 958291             # Range Header Checks
-  SecRuleRemoveById 980120             # Correlated Attack Attempt
-
-  # PDF
-  SecRuleRemoveById 920230             # Check URL encodings
-
-  # ADMIN (webdav)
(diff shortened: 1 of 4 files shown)


Iolaum:
Minor note:

I tried testing my PR with

```
$ sudo ncp-update Iolaum:dev1
INFO: updating to development branch 'Iolaum:dev1'
Downloading updates
warning: Could not find remote branch Iolaum:dev1 to clone.
fatal: Remote branch Iolaum:dev1 not found in upstream origin
No internet connectivity
```

but it looks like it's not considered a branch by github. Not sure if there's another way to test it through the ncp-update process. I 'll look into it when I can.

update.sh
```diff
@@ -186,6 +186,59 @@ EOF
 chmod +x /etc/update-motd.d/30ncp-dist-upgrade
 }
 
+# Update modsecurity config file only if user is already in buster and is used.
```
nachoparker: it will be easier to do `is_active modsecurity && run_app modsecurity`, so the `cofigure()` section is run if the user has it activated already.

Also this should be done in the `updates` folder.

nachoparker:
many thanks for this. I can merge it to `devel` so we can both test it with `ncp-update devel` once it's ready (or any other branch)

update.sh
Iolaum: Ok, I 'll refactor the code to be in the updates folder.

Iolaum:
Refactored the code per to your suggestions and re applied commits on top of current devel branch.

nachoparker:
looks good, thanks. I merged it into `devel`, so it can be tested with `ncp-update devel`

Iolaum:
@nachoparker Tested PR with `sudo ncp-upgrade devel` and it worked. A video file that wasn't syncing before (reverted my manual fix) did sync afterwards.

nachoparker:
great, many thanks!!

PR #997, opened 2019-09-10
Disable fail2ban UFW check for docker images

qrest:
Fail2ban is not present on docker images and therefore prevent a successful update.


nachoparker:
Thanks for this. Fixed

Issue #1027, opened 2019-11-14
nc-rsync-auto.sh fails for restricted remote commands


fanaz:
I want use rsync.net as a target for the nc-rsync-auto but the configuration fails with error `SSH non-interactive not properly configured`.
nc-rsync works, so the ssh access is configured properly.

nc-rsync-auto.sh tries to run the : command remotely to check if ssh is properly configured. But rsync.net provides only a restricted set of Unix commands and : is not a supported command.

nachoparker:
fixed in PR

ttoopp:
Hi, I have now the same problem with a Storage-Box from Hetzner. 
nc-rsync works, but nc-rsync-auto give me a error "SSH non-interactive not properly configured"

PR #1028, opened 2019-11-14
Extend the ssh configuration check by calling the echo command if the…

fanaz:
Fix for #1027 

To check the ssh configuration it is tested if the : command can be executed remotely. Rsync.net doesn't allow the : command, so I suggest to additionally check if the echo command can be executed remotely.

Please let me know if there's anything I can improve (this is my first contribution to an open source project).

### DIFF
--- a/bin/ncp/BACKUPS/nc-rsync-auto.sh
+++ b/bin/ncp/BACKUPS/nc-rsync-auto.sh
@@ -28,10 +28,12 @@ configure()
     return 1;
   }
 
+  # Check if the ssh access is properly configured. For this purpose the command : or echo is called remotely.
+  # If one of the commands works, the test is successful.
   [[ "$DESTINATION" =~ : ]] && {
     local NET="$( sed 's|:.*||' <<<"$DESTINATION" )"
     local SSH=( ssh -o "BatchMode=yes" -p "$PORTNUMBER" "$NET" )
-    ${SSH[@]} : || { echo "SSH non-interactive not properly configured"; return 1; }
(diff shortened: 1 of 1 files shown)


bin/ncp/BACKUPS/nc-rsync-auto.sh
```diff
@@ -28,10 +28,12 @@ configure()
     return 1;
   }
 
+  # Check if the ssh access is properly configured. For this purpose the command : or echo is called remotely.
+  # If one of the commands works, the test is successful.
   [[ "$DESTINATION" =~ : ]] && {
     local NET="$( sed 's|:.*||' <<<"$DESTINATION" )"
     local SSH=( ssh -o "BatchMode=yes" -p "$PORTNUMBER" "$NET" )
-    ${SSH[@]} : || { echo "SSH non-interactive not properly configured"; return 1; }
+    ${SSH[@]} : || ${SSH[@]} echo || { echo "SSH non-interactive not properly configured"; return 1; }
```
nachoparker:
```suggestion
    ${SSH[@]} true || { echo "SSH non-interactive not properly configured"; return 1; }
```

would this work? 

thanks for the code :)

fanaz: No, unfortunately this doesn't t work either. I think the reason is, that : and true are both bash builtins, but rsync.net supports only [a specific set of commands](https://www.rsync.net/resources/howto/remote_commands.html).

nachoparker: I see, ok then let's just leave the `echo`

nachoparker: I added it just leaving the first echo. Many thanks!

PR #1030, opened 2019-11-16
Fix for reappearing opcache bug.

Iolaum:
Fix for bug mentioned [here](https://github.com/nextcloud/nextcloudpi/issues/416#issuecomment-554265996).

If you want this in `update.sh` to stay there for subsequent versions tell me and I 'll refactor.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -191,6 +191,12 @@ EOF
 chmod +x /etc/update-motd.d/30ncp-dist-upgrade
 }
 
+# Remove redundant opcache configuration.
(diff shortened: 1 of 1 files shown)


nachoparker:
thanks,

yes, it would probably be better to leave it in `update.sh`

Iolaum:
Updated PR to have fix in `update.sh`.

Iolaum:
Closing because I just realised I should've opened the PR against devel, not master. Will re-submit

PR #1031, opened 2019-11-17
Fix for reappearing opcache bug.

Iolaum:
Fix for bug mentioned [here](https://github.com/nextcloud/nextcloudpi/issues/416#issuecomment-554265996).

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -191,6 +191,12 @@ EOF
 chmod +x /etc/update-motd.d/30ncp-dist-upgrade
 }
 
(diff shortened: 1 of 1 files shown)


Iolaum:
Got a little confused with DCO check because devel is behind master but hopefully now it's resolved.

nachoparker:
many thanks. The issue will still come back with auto-updates, but at least now it will be silenced with ncp updates.

I wonder if we should just apply this periodically instead

Issue #1035, opened 2019-12-10
Add multi factor authentication to SSH and ncp web


theCalcaholic:
**Description**  
Nowadays, MFA is considered a must-have for any kind of administrative access to web services, at least in a professional environment. Since relevant PAM modules exist, it should be easy to give users the option to enable MFA methods for SSH access as well as the admin panel (since it's based on system user authentication/pam as well).

I'm currently looking into this myself and if it's as feasible as I expect, I'll create a PR if there are no objections to the idea.

**Example**  
Some of the available pam libraries that could be used are:
- libpam-google-authenticator (which can not only be used with the official google-authenticator app, but also 3rd party apps)
- libpam-otpw
- libpam-yubico

However, I haven't evaluated any of these options in-depth yet.

nachoparker:
I like the idea!

PR #1037, opened 2019-12-14
multi factor authentication for ssh

theCalcaholic:
As proposed in #1035 I implemented an ncp app to manage multi (and single) factor authentication methods. Currently supported are:

- password only (default)
- public key authentication
- public key + password
- TOTP (google-authenticator based) + password

If multiple options are enabled, they will act as alternatives. However, if at least one mfa method is enabled, all single factor methods will be disabled automatically. 
Also, it is unfortunately impossible to use totp (or any PAM based) authentication methods with non-PAM password authentication. That makes some combinations invalid, e.g. having pubkey+pw and totp+pw both enabled.

Functionality wise I'm satisfied with the current state of the ncp app now. The only missing things that I'm aware of are documentation and localization at this point.

**TODO:**

- [x] Implement provision of ssh public key
- [x] Prevent invalid/broken combinations of parameters/authentication methods
- [x] Fix terminal version (when run via ncp-config)
- [x] ~Provide QR-Code in web interface (?)~
- [x] ~Add localization~
- [x] Test with debian buster

image

image

### DIFF
--- a/bin/ncp/SECURITY/multi-factor-authentication.sh
+++ b/bin/ncp/SECURITY/multi-factor-authentication.sh
@@ -174,13 +174,14 @@ is_active() {
 
 configure() {
 
-  local active enable_totp_and_pw enable_pubkey_and_pw enable_pubkey_only enable_pw_only reset_totp_secret
+  local active enable_totp_and_pw enable_pubkey_and_pw enable_pubkey_only enable_pw_only reset_totp_secret ssh_pubkey
   enable_totp_and_pw="$ENABLE_TOTP_AND_PASSWORD"
   enable_pubkey_and_pw="$ENABLE_PUBLIC_KEY_AND_PASSWORD"
   enable_pubkey_only="$ENABLE_PUBLIC_KEY_ONLY"
   enable_pw_only="$ENABLE_PASSWORD_ONLY"
   reset_totp_secret="$RESET_TOTP_SECRET"
   active="$ACTIVE"
+  ssh_pubkey="$(unescape "$SSH_PUBLIC_KEY")"
 
   trap 'restore' HUP INT QUIT PIPE TERM
 
@@ -234,10 +235,10 @@ configure() {
   }
 
   # Setup SSH public key if provided
-  if [[ -n "$SSH_PUBLIC_KEY" ]]
+  if [[ -n "$ssh_pubkey" ]]
   then
     echo "Setting up SSH public key..."
-    echo "$SSH_PUBLIC_KEY" > "${SSH_USER_HOME}/.ssh/authorized_keys"
+    echo "$ssh_pubkey" > "${SSH_USER_HOME}/.ssh/authorized_keys"
     chown "${SSH_USER}:" "${SSH_USER_HOME}/.ssh/authorized_keys"
   fi
 
--- a/bin/ncp/SECURITY/multi-factor-authentication.sh
+++ b/bin/ncp/SECURITY/multi-factor-authentication.sh
@@ -136,7 +136,12 @@ setup_totp_secret() {
 
   [[ "$reset_totp_secret" == "yes" ]] \
   && [[ -f "$ssh_user_home/.google_authenticator" ]] \
-  && su "$ssh_user" -c "rm '${ssh_user_home}/.google_authenticator'"
+  && {
+    echo "Deleting google authenticator configuration"
+    su "$ssh_user" -c "chmod u+w '${ssh_user_home}/.google_authenticator'"
+    su "$ssh_user" -c "rm '${ssh_user_home}/.google_authenticator'"
+  }
+
 
   if [[ "$enable_totp_and_pw" == "yes" ]] && [[ ! -f "${ssh_user_home}/.google_authenticator" ]]
   then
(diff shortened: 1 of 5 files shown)


theCalcaholic:
Somehow the ncp app is currently broken only in the terminal (`ncp-config`). I'll have to investigate that...

theCalcaholic:
@nachoparker I have now implemented the ability to supply a public key. However, it get's caught by the sanitization, because SSH public keys contain spaces (specifically at [this line](https://github.com/nextcloud/nextcloudpi/blob/master/ncp-web/ncp-launcher.php#L69)).
I know that it's hard to properly handle them in bash without security risks, but in some cases spaces are entirely valid.
I have two ideas on how to approach this, please let me know what you think about it:

1. Mark parameters in the nc-app config if they should be able to receive unsafe inputs. That way we only needed to pay attention when handling specific parameters and could apply extra caution.

2. Sanitize parameters with spaces by replacing them (e.g. with html special chars). That way, if a script needs access to the unescaped variable it could just reverse the process with a simple search & replace (maybe a function for that should be provided via the library.sh).

theCalcaholic:
I have taken the liberty to implement the 3rd option as an example. If you are not fine with it, it can always be reverted.

bin/ncp/SECURITY/multi-factor-authentication.sh
```diff
@@ -0,0 +1,286 @@
...
+configure() {
+
+  local active enable_totp_and_pw enable_pubkey_and_pw enable_pubkey_only enable_pw_only reset_totp_secret ssh_pubkey
+  enable_totp_and_pw="$ENABLE_TOTP_AND_PASSWORD"
+  enable_pubkey_and_pw="$ENABLE_PUBLIC_KEY_AND_PASSWORD"
+  enable_pubkey_only="$ENABLE_PUBLIC_KEY_ONLY"
+  enable_pw_only="$ENABLE_PASSWORD_ONLY"
+  reset_totp_secret="$RESET_TOTP_SECRET"
+  active="yes"
+  ssh_pubkey="$(unescape "$SSH_PUBLIC_KEY")"
```
nachoparker: if we want to allow spaces, it sounds like we should use bash arrays

theCalcaholic: That doesn't sound right to me. IMO, bash arrays are rather unintuitive for this purpose, because the array fields have no meaning (it's not a space separated set of strings, but a string containing spaces).

theCalcaholic: Why would you prefer bash arrays here?

nachoparker: what is the issue exactly? is it the sanitizing issue above? if the issue is ncp-config getting confused by spaces we would have to see if it makes sense to use arrays. It is probably a bad idea in terms of sanitizing/security to allow spaces as an input but we would have to take a close look

theCalcaholic: The issue is that many values for ncp apps might validly contain spaces (e. g. paths or, like in this case, an ssh pubkey).
Therefore we need a way to provide unsanitized input values.
This might make sense for more than spaces (although not required in this case).

Have you had a look at my change to ncp-launcher.php?

nachoparker: The problem with allowing spaces (in any form) is that an attacker could potentially craft a command with arguments if he is able to bypass other restrictions.

If we disallow spaces, at least all the attacker can do at that point is to run a command with arguments which would hopefully be much less useful for him.

How about uploading the pubkey, or having a special field that saves the unsanitized value to a file?

theCalcaholic: That kind of attack is not possible if unsafe values are handled the way I implemented it.

- Spaces are being replaced by "%SPACE%" in the input values.
- They must be explicitly unescaped in bash in order to get the 'unsafe' value

If you are aware that a value might contain spaces, you can easily avoid that kind of argument injection.

In this particular case, a file upload would work as well - however, people are complaining about missing support for spaces (e. g. in paths) since forever, so that would solve a lot more cases than just ssh public key provision.

nachoparker: Mmm I don't see a way where spaces would be an issue in this case. Makes me a bit uneasy, but if we are extra careful looks ok. Hopefully we are not missing anything.

theCalcaholic: Well, if you prefer another option - I can work with any of the mentioned ideas for this use case :)
But personally, I'd prefer this implementation (my second pick would be bash arrays, since it provides the same functionality but is harder to implement and less intuitive).

nachoparker: like I said, I don't see how in this case your implementation could be insecure. I think it is fine if you also triple check that there is no way to inject commands.

theCalcaholic: Alright. :) 

I was just saying that, because my implementation includes changes to ncp-launcher.php and library.sh, meaning it will be available to other apps in the future.

nachoparker: great, let me know when you are happy with this and I'll give it a deeper review

theCalcaholic: I am now :)

theCalcaholic:
Hm, maybe it would be a good idea to add a few more text fields for ssh public keys...

I'll implement that in a bit

theCalcaholic:
@nachoparker Alright, it's feature complete for now

nachoparker:
Thanks! I will take a look soon, but I don't know when since I have family visits for the following weeks.

theCalcaholic:
Sure, take your time :)

Issue #1038, opened 2019-12-16
LetsEncrypt: Second Domain: Invaid character "_"


langfingaz:
Thanks for the recent support to setup a second domain through ncp-config (#1025). By default if one runs the letsencrypt ncp-script, a `_` is preselected for both domains.

I just changed the first `_` to my desired domain and left the second domain as `_`. I ran into the following error:

**Error message**

<details>
<summary>Cannot issue for "_": Domain name contains an invalid character</summary>

```
Running letsencrypt
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Plugins selected: Authenticator webroot, Installer None
Obtaining a new certificate
An unexpected error occurred:
Error creating new order :: Cannot issue for "_": Domain name contains an invalid character
Please see the logfiles in /var/log/letsencrypt for more details.
Done. Press any key...

```

</details>

**System information**

<details>
<summary>NextCloudPi diagnostics</summary>

```

NextCloudPi version  v1.19.1
NextCloudPi image    NextCloudPi_docker_07-20-19
distribution         Debian GNU/Linux 10 \n \l
automount            no
USB devices          none
datadir              /data/nextcloud/data
data in SD           yes
data filesystem      btrfs
data disk usage      108G/1.9T
rootfs usage         11G/117G
swapfile             none
dbdir                /data/database
Nextcloud check      ok
Nextcloud version    17.0.1.1
HTTPD service        up
PHP service          up
MariaDB service      up
Redis service        up
Postfix service      up
internet check       ok
port check 80        open
port check 443       open
IP                   ***REMOVED SENSITIVE VALUE***
gateway              ***REMOVED SENSITIVE VALUE***
interface            eth0
certificates         ***REMOVED SENSITIVE VALUE***
NAT loopback         no
uptime               7days

```
</details>

langfingaz:
I guess the strings "" and "_" as domain should be ignored while iterating over the given domains in https://github.com/nextcloud/nextcloudpi/edit/master/bin/ncp/NETWORKING/letsencrypt.sh

theCalcaholic:
I'd say "_" is just a bad default character. ncp supports suggestions for text fields, which, imo,  should be used instead.

langfingaz:
I just pulled the latest x86 nextcloudpi docker container and tried to reproduce the behavior with the prefilled `"_"` on a fresh instance:

[...]

PR #1039, opened 2019-12-16
Update letsencrypt.sh

langfingaz:
Fix "_" invalid domain error as described here: https://github.com/nextcloud/nextcloudpi/issues/1038


nachoparker:
Many thanks.

While this would work I think it would be worthwile to fix the design issue, since we can probably avoid the `_` altogether.

To me that's the ideal solution, if somebody like yourself or @theCalcaholic wanted to implement the change.

We can reopen this if we don't find an alternative.

Issue #1047, opened 2020-01-02
samba release only for one/some nextcloud user


hurrikan77:
**Description**  
Actually it is possible to activate the sambe sharing function for all users which are created within the nextcloud instance. I would like to have only sambe release for selected user/users.

**Example**  
Nextcloud User 1
Nextcloud User 2
Nextcloud User 3

After selecting the navigation link I would like to have an selection field for which user/users I would like to aktive the samba sharing function.

theCalcaholic:
That would be a nice feature, but is hard to implement with the current ncp-app architecture.
Dynamic configuration items are currently not supported (they are read from static text files atm).

So... good idea, but probably too much effort right now I'd expect.

hurrikan77:
If you create for each user an user group "e.g. samba" within nextcloud.
Is there an possibility to add to the loop for creating samba users an request if the user is member of the "samba" User group.
If yes create the samba sharing, if not skip the user.

theCalcaholic:
Not a bad idea. A "filter_by_group" option might work

theCalcaholic:
@hurrikan77 Out of curiosity: What's your use case for that feature? Why do you want to enable samba only for a portion of your users?

nachoparker:
current support for the feature is very basic. All samba clients share the same password...

hurrikan77:
@theCalcaholic 
I have done a little bit of rasperry pi "recycling" during the holidays. 
I set up some Volumio instances to play music in different rooms. Base is an music depot which I host in my nextcloud instance. I created an new user for that and shared the files of this user via samba with each instance. So I could prevent to copy the whole data on each of the pi's.

That is the case why I asked for only one/some user. For all other users ist not nessesary to have samba Access to there files.

hurrikan77:
@nachoparker 
And thats one reason why I want to reduce the number of open shares to a minimum.

nachoparker:
the PR @theCalcaholic implemented is a step forward. Still no password per user or anything like that...

Thanks @theCalcaholic

PR #1048, opened 2020-01-03
SAMBA - filter by group option

theCalcaholic:
Implements #1047

At this point this is a quickly thrown together (untested!) implementation (10min effort :P ).
I'll iron it out in the coming days if Nacho likes the approach.

### DIFF
--- a/bin/ncp/NETWORKING/samba.sh
+++ b/bin/ncp/NETWORKING/samba.sh
@@ -69,12 +69,12 @@ EOF
 
   for user in ${USERS[@]}; do
     # Exclude users not matching group filter (if enabled)
-    [[ -n "$FILTER_BY_GROUP" ]] \
-    && [[ -z "$(ncc user:info "$user" --output=json | jq ".groups[] | select( . == \"${FILTER_BY_GROUP}\" )")" ]] \
-    || {
+    if [[ -n "$FILTER_BY_GROUP" ]] \
+    && [[ -z "$(ncc user:info "$user" --output=json | jq ".groups[] | select( . == \"${FILTER_BY_GROUP}\" )")" ]]
+    then
       echo "Omitting user $user (not in group ${FILTER_BY_GROUP}...";
       continue;
-    }
+    fi
 
     echo "adding SAMBA share for user $user"
     local DIR="$DATADIR/$user/files"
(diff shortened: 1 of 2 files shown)


etc/ncp-config.d/samba.cfg
```diff
@@ -17,6 +17,11 @@
       "name": "Password",
       "value": "ownyourbits",
       "type": "password"
+    },
+    {
+      "id": "FILTER_BY_GROUP",
+      "name": "Filter by NC group:",
+      "value": ""
```
nachoparker: maybe a suggest value `optional_group` and maybe some explanation in the description, but in general looks good to me.

theCalcaholic: Sure. Is `OPTIONAL_GROUP_FILTER` okay?

nachoparker: sounds good to me

theCalcaholic: Ah, I misunderstood you earlier. I've updated the description and added a suggest value

theCalcaholic: @nachoparker Let me know if my recent change implements what you had in mind (once you get to it - there's no hurry).

bin/ncp/NETWORKING/samba.sh
```diff
@@ -68,6 +68,14 @@ EOF
   done < <( ls -d "$DATADIR"/*/files )
 
   for user in ${USERS[@]}; do
+    # Exclude users not matching group filter (if enabled)
+    if [[ -n "$FILTER_BY_GROUP" ]] \
+    && [[ -z "$(ncc user:info "$user" --output=json | jq ".groups[] | select( . == \"${FILTER_BY_GROUP}\" )")" ]]
+    then
+      echo "Omitting user $user (not in group ${FILTER_BY_GROUP}...";
```
nachoparker:
```suggestion
      echo "Omitting user $user (not in group ${FILTER_BY_GROUP})...";
```

etc/ncp-config.d/samba.cfg
```diff
@@ -17,6 +17,12 @@
       "name": "Password",
       "value": "ownyourbits",
       "type": "password"
+    },
+    {
+      "id": "FILTER_BY_GROUP",
+      "name": "Filter by NC group:",
```
nachoparker:
```suggestion
      "name": "Filter by NC group",
```

nachoparker:
merged to `devel`. Is it tested by you? good to merge to `master`?

theCalcaholic:
I did not test it yet, because I wanted to hear your opinion on the general idea first!
I'll try to do that tomorrow (I'm on a tight schedule this week, because it's my wife's birthday :P)

PR #1059, opened 2020-01-18
Add user only if it does not exist.

budulinek:
If user already exists, useradd ends with error and terminates the whole instalation process.


nachoparker:
thanks, I appreciate it

PR #1062, opened 2020-01-21
[WIP] Don't save passwords

theCalcaholic:
Well, actually "delete passwords when not needed anymore".

## Description
We should try to avoid storing clear text passwords as well as possible. However, with the current architecture, it's not possible to not at least temporarily save them to the configuration file.
The best thing we can do, is to delete the passwords from the configuration when they aren't needed anymore (usually after being applied).

This PR adds a function in library.sh which allows an ncp-app to do just that and implements it in nc-passwd.sh as a reference.

Let me know if you consider this a good approach or if you see any issues with it.

**EDIT:** Of course this would take away the ability to show the current password in the respective ncp-app. But in my book, that's a feature, not a bug. :wink:


theCalcaholic:
Btw, @nachoparker, do you prefer PRs to the master or the devel branch? Because I noticed recently that devel was behind master, so I wasn't sure how much sense it makes to create a PR for devel...

ncp-web-neo/.gitignore
```diff
@@ -0,0 +1,21 @@
+.DS_Store
```
nachoparker: accidentally saved file

bin/ncp/CONFIG/nc-passwd.sh
```diff
@@ -8,10 +8,16 @@
 # More at: https://ownyourbits.com
 #
 
-
+cleanup()
```
nachoparker: let's do this for every ncp-app

ncp-web-neo/.gitignore
theCalcaholic: Not _really_ accidentally - but yeah, I should remove that again :D

bin/ncp/CONFIG/nc-passwd.sh
theCalcaholic: Do you think I can implement this as an opt-out instead of an opt-in (or not optional at all) without breaking things?
If so I'd include it in `library.sh/run_app_unsafe` instead

nachoparker:
testing this

```
diff --git a/bin/ncp/CONFIG/nc-passwd.sh b/bin/ncp/CONFIG/nc-passwd.sh                                                                                                                                                                                                                                                [25/1597]
index 52a3813..2aed52f 100644                                                                                                                                                                                                                                                                                                  
--- a/bin/ncp/CONFIG/nc-passwd.sh                                                                                                                                                                                                                                                                                              
+++ b/bin/ncp/CONFIG/nc-passwd.sh                                                                                                                                                                                                                                                                                              
@@ -8,16 +8,8 @@                                                                                                                                                                                                                                                                                                               
 # More at: https://ownyourbits.com                                                                                                                                                                                                                                                                                            
 #                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                               
-cleanup()                                                                                                                                                                                                                                                                                                                     
-{                                                                                                                                                                                                                                                                                                                             
-  # Remove cleartext passwords from configuration                                                                                                                                                                                                                                                                             
-  clear_all_passwords "$0"                                                                                                                                                                                                                                                                                                    
-}                                                                                                                                                                                                                                                                                                                             
-                                                                                                                                                                                                                                                                                                                              
 configure()                                                                                                                                                                                                                                                                                                                   
 {                                                                                                                                                                                                                                                                                                                             
-  trap cleanup EXIT SIGHUP SIGABRT SIGINT                                                                                                                                                                                                                                                                                     
-                                                                                                                                                                                                                                                                                                                              
   # update password                                                                                                                                                                                                                                                                                                           
   echo -e "$PASSWORD\n$CONFIRM" | passwd ncp &>/dev/null && \                                                                                                                                                                                                                                                                 
     echo "password updated successfully" || \                                                                                                                                                                                                                                                                                 
@@ -36,7 +28,6 @@ configure()                                                                                                                                                                                                                                                                                                  
   a2ensite  ncp nextcloud                                                                                                                                                                                                                                                                                                     
   a2dissite ncp-activation                                                                                                                                                                                                                                                                                                    
   bash -c "sleep 1.5 && service apache2 reload" &>/dev/null &                                                                                                                                                                                                                                                                 
-                                                                                                                                                                                                                                                                                                                              
 }
  
 install() { :; }
diff --git a/etc/library.sh b/etc/library.sh
index eb2c070..e9b5651 100644
--- a/etc/library.sh
+++ b/etc/library.sh
@@ -150,6 +150,7 @@ function run_app_unsafe()
  
   echo "" >> $log
  
+  clear_password_fields "$cfg_file"
   return "$ret"
 }
  
@@ -278,14 +279,10 @@ function check_distro()
   return 1
 }
  
-function clear_all_passwords()
+function clear_password_fields()
 {
-  local script=$1
-  local ncp_app="$(basename "$script" .sh)"
-  local cfg_file="$CFGDIR/$ncp_app.cfg"
-  local log=/var/log/ncp.log
-
-  cfg="$(jq '.params[] | select(.type == "password") | .value = ""' <"$cfg_file")"
+  local cfg_file="$1"
+  local cfg="$(jq '.params[] | select(.type == "password") | .value = ""' < "$cfg_file")"
   echo "$cfg">"$cfg_file"
 }
```

bin/ncp/CONFIG/nc-passwd.sh
theCalcaholic: Btw, will trap even trigger if a function returns? :thinking:

nachoparker: look at my changes below, I am building them. I implemented it in `run_app_unsafe`

https://github.com/nextcloud/nextcloudpi/pull/1062#issuecomment-576725577

theCalcaholic:
@nachoparker I commited changes reflecting this

nachoparker:
devel always, it shouldn't be behind master except in very weird situations

nachoparker:
this works

```
commit a26cfa39f498ab3986403a68e6544f3bd217087c (HEAD -> theCalcaholic-feature/dont-save-passwords, origin/theCalcaholic-feature/dont-save-passwords)
Author: nachoparker <nacho@ownyourbits.com>
Date:   Tue Jan 21 20:03:23 2020 -0700

    fix
    
    Signed-off-by: nachoparker <nacho@ownyourbits.com>

diff --git a/etc/library.sh b/etc/library.sh
index e9b5651..461b2e5 100644
--- a/etc/library.sh
+++ b/etc/library.sh
@@ -282,8 +282,15 @@ function check_distro()
 function clear_password_fields()
 {
   local cfg_file="$1"
-  local cfg="$(jq '.params[] | select(.type == "password") | .value = ""' < "$cfg_file")"
-  echo "$cfg">"$cfg_file"
+  local cfg="$(cat "$cfg_file")"
+  local len="$(jq '.params | length' <<<"$cfg")"
+  for (( i = 0 ; i < len ; i++ )); do
+    local type="$(jq -r ".params[$i].type"  <<<"$cfg")"
+    local val="$( jq -r ".params[$i].value" <<<"$cfg")"
+    [[ "$type" == "password" ]] && val=""
+    cfg="$(jq -r ".params[$i].value=\"$val\"" <<<"$cfg")"
+  done
+  echo "$cfg" > "$cfg_file"
 }
 
 function apt_install()
```

nachoparker:
Done it myself https://github.com/nextcloud/nextcloudpi/commit/4a99207a7427949202c449eb4aa78da5d9ae2d22

thanks!

PR #1073, opened 2020-01-31
Add more info to description of nc-backup.cnf

ovpc:
What seems obvious to us, may be unclear for 1st timers ;-)

People ask about this and backups in general.


nachoparker:
I like it, thanks. I would omit the snapshots part because it can be confusing. You still need to somehow transfer the snapshot to the new drive (maybe btrfs-sync), since that data is not included in the dataless backup and from the description it might seem like it is.

ovpc:
Agree, could be confusing.

PR #1079, opened 2020-02-19
Exclude database directory from COW (on btrfs partitions)

theCalcaholic:
While reading through nc-database.sh, I thought that it might make sense to set the NO_COW flag on the db directory with chattr +C (because cow is not useful on large binary files and can heavily reduce performance).
This PR does precisely that. :)

I'm not entirely sure what happens when creating a btrfs snapshot, though. @nachoparker Do you think it might cause any issues?

### DIFF
--- a/bin/ncp/CONFIG/nc-database.sh
+++ b/bin/ncp/CONFIG/nc-database.sh
@@ -42,8 +42,8 @@ configure()
   
   # Set NOCOW flag on DBDIR if it is located on a btrfs file system
   mkdir -p "$DBDIR"
-  chown --reference="$SRCDIR"
-  chmod --reference="$SRCDIR"
+  chown --reference="$SRCDIR" "$DBDIR"
+  chmod --reference="$SRCDIR" "$DBDIR"
 
   grep -q -e btrfs <<<"$FSTYPE" && chattr +C "$DBDIR"
 
--- a/bin/ncp/CONFIG/nc-database.sh
+++ b/bin/ncp/CONFIG/nc-database.sh
@@ -31,6 +31,12 @@ configure()
   local BASEDIR=$( dirname "$DBDIR" )
   mkdir -p "$BASEDIR"
 
+  local FSTYPE="$(df -T /home/tknoeppl | cut -d " " -f2 | tail -n 1)"
+  if [[ $FSTYPE == "btrfs" ]]
+  then
+    chattr +C "$BASEDIR"
+  fi
+
   grep -q -e ext -e btrfs <( stat -fc%T "$BASEDIR" ) || { echo -e "Only ext/btrfs filesystems can hold the data directory"; return 1; }
   
   sudo -u mysql test -x "$BASEDIR" || { echo -e "ERROR: the user mysql does not have access permissions over $BASEDIR"; return 1; }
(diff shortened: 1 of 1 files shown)


bin/ncp/CONFIG/nc-database.sh
```diff
@@ -31,7 +31,11 @@ configure()
   local BASEDIR=$( dirname "$DBDIR" )
   mkdir -p "$BASEDIR"
 
-  grep -q -e ext -e btrfs <( stat -fc%T "$BASEDIR" ) || { echo -e "Only ext/btrfs filesystems can hold the data directory"; return 1; }
+  local FSTYPE="$( stat -fc%T "$BASEDIR" )"
+
+  grep -q -e btrfs <<<"$BASEDIR" && chattr +C "$BASEDIR"
+
+  grep -q -e ext -e btrfs <<<"$BASEDIR" || { echo -e "Only ext/btrfs filesystems can hold the data directory"; return 1; }
```
nachoparker:
```suggestion
  grep -q -e ext -e btrfs <<<"$FSTYPE" || { echo -e "Only ext/btrfs filesystems can hold the data directory"; return 1; }
```

bin/ncp/CONFIG/nc-database.sh
```diff
@@ -31,7 +31,11 @@ configure()
   local BASEDIR=$( dirname "$DBDIR" )
   mkdir -p "$BASEDIR"
 
-  grep -q -e ext -e btrfs <( stat -fc%T "$BASEDIR" ) || { echo -e "Only ext/btrfs filesystems can hold the data directory"; return 1; }
+  local FSTYPE="$( stat -fc%T "$BASEDIR" )"
+
+  grep -q -e btrfs <<<"$BASEDIR" && chattr +C "$BASEDIR"
```
nachoparker:
```suggestion
  grep -q -e btrfs <<<"$FSTYPE" && chattr +C "$BASEDIR"
```

theCalcaholic: :man_facepalming:

theCalcaholic:
@nachoparker Finally got around to testing this PR.
I had to fix up one thing, namely, the directory attributes got overriden by `mv`, so I had to create the directory beforehand and mv only the files/subdirectories of the DB directory.

I'm not totally happy about using bash's path extension for this (especially since I need to set `shopt -s dotglob` in order to match potential hidden files), but I disliked the alternative (for loop over the directory contents) even more, so I went with it.

If you have a better idea, please let me know :)

As far as I'm concerned, this PR is ready to be merged now (tested on a non-docker installation, but shouldn't matter in this case).

theCalcaholic:
Btw, I found this link on the behavior of nocow files for snapshots:

https://unix.stackexchange.com/questions/285763/taking-snapshots-of-a-btrfs-volume-mounted-with-nodatacow#286052

Turns out, that

1. Snapshots still work and include nocow files

2. The first write to each block of a nocow file after a snapshot was made will be cow - after that, the nocow flag will be effective again.

theCalcaholic:
@nachoparker If that's okay with you, I could as well just merge the PR. Just need your okay :)

nachoparker:
let's do it. Thanks :)

PR #1021, opened 2019-10-28
fixing file permission for redis config

mad73923:
I'm running ncp in docker.
This change fixed my problems with redis after updating.

Changing the redis.conf file leads to a permission change. This has to be reverted by the added command.

We also might have to change 
`bin/ncp/CONFIG/nc-nextcloud.sh`
since the file is adapted there as well.

fixes #1019

PR #1081, opened 2020-02-22
add missing chown in install function for redis config, fixes 1019

mad73923:
in #1021, I already mentioned we might have to change this. But I forgot about it.
Now, as I tried to install a new instance of nextcloudpi, this issue came up again :angry:


mad73923:
could you please also update the docker armhf image on dockerhub?

nachoparker:
added the line manually... thank you!!

Issue #1120, opened 2020-04-06
Include php-gmp in all docker builds


StableNarwhal:
At least on the current armhf build (v1.24.0) php-gmp isn't included, leading to errors when installing specific Nextcloud apps, e.g. **Bookmarks**:

`App “Bookmarks” cannot be installed because the following dependencies are not fulfilled: The library gmp is not available.`

The manual fix is to connect to the docker container, e.g. via `docker exec -it nextcloudpi /bin/bash
`, installing it manually via `sudo apt install php-gmp` and restarting the container afterwards via `docker restart nextcloudpi`

There are some reports that this isn't an issue for other people on other builds, if this is indeed missing for the armhf and maybe other builds it should be included in the future.

nachoparker:
the image was bad

PR #1121, opened 2020-04-06
Includes php-gmp in the installed packages

StableNarwhal:
This should fix https://github.com/nextcloud/nextcloudpi/issues/1120.


nachoparker:
the image was bad, thanks for the help!

PR #1133, opened 2020-05-05
Don't fail ncp-backup-auto

llucax:
Make sure the script don't return an exit status different from 0 so the
cron job doesn't send e-mails on failure, as notifications are being
delivered as Nextcloud notifications. Also redirect all output to a log
file, so in case of failure it is possible to review what happened.

This also makes the notification a bit more informative, as it says
which script failed.

The log files is overwritten on each run. This should not be a problem
since backups are run more frequently than once per day, so there should
be a fair amount of time to see the backup logs.

### DIFF
--- a/bin/ncp/BACKUPS/nc-backup-auto.sh
+++ b/bin/ncp/BACKUPS/nc-backup-auto.sh
@@ -19,17 +19,28 @@ configure()
   cat > /usr/local/bin/ncp-backup-auto <<EOF
 #!/bin/bash
 source /usr/local/etc/library.sh
-[ -x /usr/local/bin/ncp-backup-auto-before ] && /usr/local/bin/ncp-backup-auto-before
+failed=
+run_script()
+{
+        if [ -x /usr/local/bin/ncp-backup-auto-\$1 ]
+        then
+                /usr/local/bin/ncp-backup-auto-\$1 || failed="\$failed\${failed:+, } \$1"
+        fi
+}
+
+run_script before
 /usr/local/bin/ncc maintenance:mode --on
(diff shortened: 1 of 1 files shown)


nachoparker:
thanks for the PR, overall looks good, let me know when it's finished

llucax:
Hi, you are welcome, thanks for putting nexcloudpi together :)

From my side I'd say it is finished, but let me know if I missing anything.

nachoparker:
can we output the log to `ncp.log`? the reason is that we have proper log rotation set up for that one

bin/ncp/BACKUPS/nc-backup-auto.sh
```diff
@@ -19,17 +19,28 @@ configure()
...
+run_script after
+if [[ -n "\$failed" ]]
+then
+  notify_admin "Auto-backup failed" "The \$failed backup script(s) failed"
+fi
 EOF
   chmod +x /usr/local/bin/ncp-backup-auto
 
-  echo "0  3  */${BACKUPDAYS}  *  *  root  /usr/local/bin/ncp-backup-auto" > /etc/cron.d/ncp-backup-auto
+  echo "0  3  */${BACKUPDAYS}  *  *  root  /usr/local/bin/ncp-backup-auto > /var/log/ncp-backup-auto.log 2>&1" > /etc/cron.d/ncp-backup-auto
```
nachoparker, line 43:
```suggestion
  echo "0  3  */${BACKUPDAYS}  *  *  root  /usr/local/bin/ncp-backup-auto > /var/log/ncp.log 2>&1" > /etc/cron.d/ncp-backup-auto
```

nachoparker:
thanks!

PR #1134, opened 2020-05-05
Don't fail ncp-autoupdate-apps cron if no updates

llucax:
If the `[[ "\$APPS" != "" ]]` check fails, the shell will use it's exit status as the whole script exit status, making the script return 1. Then cron thinks this job failed and sends an e-mail notification, which is not really providing any useful information as it only means nothing was updated.


nachoparker:
good catch. Fixed, thanks!

nachoparker:
https://github.com/nextcloud/nextcloudpi/commit/81fcd1483ac60677742aadd5656aded3009ce18f

PR #1138, opened 2020-05-17
Feature/encrypted backups

danploeger:
Hello, I felt that NCP was lacking a proper backup solution with encrypted containers, which can be synchronized to an untrusted remote destination.

For this reason I implemented the scripts `nc-rsync-encrypted` and `nc-rsync-encrypted-auto` during the last days. The scripts are tested, but not exhaustively. The features are: 

- A full encrypted backup can be sent to a remote destination one-time with `nc-rsync-encrypted`
- Encrypted backups can be sent to a remote destination on a regular basis with `nc-rsync-encrypted-auto`
- Either NC data or database or both can be backed up
- The encrypted backups use duplicity via rsync; an existing GPG key on the NCP server is required
- The auto backup is configured with full sync and incremental sync periodicities. The incremental syncs are faster, but increase the backup size at the remote destination with each backup. On the other hand, the full syncs do not keep outdated/dispensable data on the remote destination
- The one-time backup doesn't store any confidential information (due to GPG asymmetric encryption). This limits the backup to single backups per remote destination, because the metadata of previous backups can't be decrypted.
- The auto backup stores the GPG passphrase in a hidden file in the root home directory, only accessible by the NCP root user. This allows for regular full and incremental backups
- Note the slightly different destination path syntax on the web interface compared to `rc-rsync`. Instead of rsync access via remote shell, the destination is accessed via the rsync daemon to allow duplicity usage
- Suggestion for the docs.nextcloudpi.com configuration reference:
```
Step 1: follow the connection setup reference for nc-rsync.

Step 2: on the NCP server, generate a GPG key pair with the following command:

gpg --gen-key

During the key generation process you are asked to enter a passphrase. Note it down at a safe place. This is required if you ever want to restore your backup. The public key is stored at /root/.gnupg/, to get a list of all keys use the following command:

gpg --list-keys

The list shows the GPG key ID, a long hexadecimal hash value. Paste this value to the respective field of nc-rsync-encrypted or nc-rsync-encrypted-auto. 
```
- Additional paragraph for the docs.nextcloudpi.com configuration reference of `nc-rsync-encrypted-auto`:
```
In nc-rsync-encrypted-auto, enter your passphrase to the GPG passphrase field. The auto backup stores the GPG passphrase in a hidden file in the root home directory, only accessible by the NCP root user. This allows for regular full and incremental backups. Choose full sync and incremental sync periodicities. The incremental syncs are faster, but increase the backup size at the remote destination with each backup. On the other hand, the full syncs do not keep outdated/dispensable data on the remote destination.
```
- Future work: implement the restore functionality, add localisation, add additional usage checks for unexpected duplicity behavior.

### DIFF
--- a/bin/ncp/BACKUPS/nc-rsync-encrypted-auto.sh
+++ b/bin/ncp/BACKUPS/nc-rsync-encrypted-auto.sh
@@ -18,9 +18,11 @@ install()
 
 configure()
 {
-  [[ $ACTIVE != "yes" ]] && { 
+  [[ $ACTIVE != "yes" || ( $INCLUDEDB != "yes" && $INCLUDEDATA != "yes" ) ]] && { 
 	rm -f /root/.passphrase
     rm -f /etc/cron.d/ncp-rsync-encrypted-auto
+	rm -f /usr/local/bin/ncp-rsync-encrypted-full
+	rm -f /usr/local/bin/ncp-rsync-encrypted-incr
     echo "automatic encrypted rsync disabled"
     return 0
   }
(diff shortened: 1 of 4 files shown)


theCalcaholic:
Hey, thanks for the PR! That's a really cool feature.

Do you think it would be possible to merge nc-rsync-encrypted with nc-rsync? If possible, a mere 'encrypt backups' checkbox would be more user friendly than two distinct scripts.

nachoparker:
Interesting, thanks.

My favorite choice for incremental + encrypted is to use BTRFS snapshots. We already use those, only that now I am working on encrypting them as well, and that should be very efficient. We'll see how it goes.

Well, this is for the data. For the backups... still would have to think about something. Borg would be a big change but it's not out of the question.

PR #1144, opened 2020-06-13
Make ncp-autoupdate cronjob write to the log

llucax:
Cronjobs should not have any output, unless the cronjob fails, otherwise an e-mail notification is sent with the output. This cronjob in particular is also reporting results via Nextcloud notification, so having an e-mail with logs saying the same as the notification is not particularly useful. Instead we just redirect all output to the /var/log/ncp.log so the user can debug in case of problems and doesn't receive spurious e-mails when all works fine.


nachoparker:
looks good, thanks!

Issue #777, opened 2019-01-16
The .htaccess file is not working.


LukasWerfel:
After a freshly installed ncp instance, I just deactivated HTTPS (its accessible via nginx on another server through VPN, and access to that server is HTTPS) and I get this. My files are exposed. How can I fix this?

> Your data directory and files are probably accessible from the Internet. The .htaccess file is not working. It is strongly recommended that you configure your web server so that the data directory is no longer accessible, or move the data directory outside the web server document root.

I fixed it for now by blocking that route on nginx, so I am not sure if there might be a way around this measure.

nachoparker:
nginx? NCP doesn't use nginx.

Please, use the forum for questions and provide information from `sudo ncp-report`

cheers

AvverbioPronome:
@LukasWerfel I too have an nginx reverse proxy in front of the ncp apache. How did you solve this issue?

AvverbioPronome:
Ok, found the issue. There is no `AllowOverride All` in the http virtualhost: https://github.com/nextcloud/nextcloudpi/blob/4a51c1f5e8a33e670a63c4c82f7eae7115b86980/bin/ncp/CONFIG/nc-nextcloud.sh#L175

nachoparker:
`AllowOverride All` is the default.

http://httpd.apache.org/docs/2.2/mod/core.html#AllowOverride

AvverbioPronome:
> `AllowOverride All` is the default.

Not on Debian: https://salsa.debian.org/apache-team/apache2/-/blob/master/debian/config-dir/apache2.conf.in#L153

nachoparker:
Oh I see, we do have `AllowOverride All` in `nextcloud.conf`, but if you disable HTTP->HTTPS redirection then it wouldn't apply, good catch.

PR #1158, opened 2020-07-24
Reverse proxy and hsts improvements

AvverbioPronome:
This should solve issue #777 and use more sensible defaults for the hsts header (that is ignored anyway when not served over https)

### DIFF
--- a/bin/ncp/CONFIG/nc-nextcloud.sh
+++ b/bin/ncp/CONFIG/nc-nextcloud.sh
@@ -210,6 +210,14 @@ EOF
     RewriteCond %{HTTPS} !=on
     RewriteRule ^/?(.*) https://%{SERVER_NAME}/$1 [R,L]
   </IfModule>
+  <Directory /var/www/nextcloud/>
+    Options +FollowSymlinks
+    AllowOverride All
+    <IfModule mod_dav.c>
+      Dav off
+    </IfModule>
+    LimitRequestBody 0
+  </Directory>
 </VirtualHost>
 EOF
 
--- a/bin/ncp/CONFIG/nc-nextcloud.sh
+++ b/bin/ncp/CONFIG/nc-nextcloud.sh
@@ -196,8 +196,6 @@ EOF
   <IfModule mod_headers.c>
     Header always set Strict-Transport-Security "max-age=15768000; includeSubDomains"
   </IfModule>
-EOF
-
 </IfModule>
 EOF
   a2ensite nextcloud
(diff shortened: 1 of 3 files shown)


nachoparker:
Also, we should apply these changes for existing users through the `nc-update` mechanism. See the `updates/` folder.

AvverbioPronome:
> Note that by removing HSTS from the global settings you are also removing it from ncp-web

I didn't think about this but it's probably not an issue, hsts is (sub)domain-wide. Where's the ncp-web virtualhost defined?

> Also, we should apply these changes for existing users through the `nc-update` mechanism. See the `updates/` folder.

Replace or edit? I can't imagine a safe way of doing that.

AvverbioPronome:
I've been looking at the source code to understand where/when to apply hsts headers to ncp-web, and I realised that we can't apply them at the very beginning with a self signed certificate, we need to wait for a Let's Encrypt certificate to be obtained and working. 

Ideas?

(I also thought we can probably create an `/etc/apache2/snippets/ directory for all the reused configurations across the virtualhosts, but I wanted to hear from you before messing with it.)

(Another idea more could be to move from certbot to acme.sh or bacme -- which are shell scripts instead of full-fledged python applications)

nachoparker:
> (Another idea more could be to move from certbot to acme.sh or bacme -- which are shell scripts instead of full-fledged python applications)

We have been toying with this idea, since you can use non standard ports, but certbot is a defacto standard and it works well, so we didn't see a reason to go through the hassle.

nachoparker:
> (I also thought we can probably create an /etc/apache2/snippets/ directory for all the reused configurations across the virtualhosts, but I wanted to hear from you before messing with it.)

I would be fine with that.

> I've been looking at the source code to understand where/when to apply hsts headers to ncp-web, and I realised that we can't apply them at the very beginning with a self signed certificate, we need to wait for a Let's Encrypt certificate to be obtained and working.

I think that if we just leave them where they were they apply to every SSL connection, no matter if there are cert changes.

nachoparker:
ncp-web vhost -> https://github.com/nextcloud/nextcloudpi/blob/master/ncp.sh

AvverbioPronome:
> I think that if we just leave them where they were they apply to every SSL connection, no matter if there are cert changes.

Yes and no:
- right now the header gets sent on http too, even if it's meaningless -- but this also means that if I use a reverse proxy I send **two** hsts headers: my own from the tls terminator, and ncp's one.
- a possible issue: hsts *and* a self-signed certificate pretty much means disallowing any connection. :D

AvverbioPronome:
> We have been toying with this idea, since you can use non standard ports, but certbot is a defacto standard and it works well, so we didn't see a reason to go through the hassle.

Certbot can use different ports as well, it's not about features (ok, it is if we want to use elliptic curve certificates), it's more about reducing dependencies.

AvverbioPronome:
I think as far as improvements/new features are concerned, this pull request is done.

I/we should get it ready to be merged. You say the `nc-update` mechanism?

nachoparker:
applied, many thanks!

PR #1170, opened 2020-08-29
Fix the style of the language selection dropdown.

powerflo:
If the language selection is active the background color as well as the color was white and therefore the language options were unreadable.
Only Chrome is affected (tested with Chromium Version 84.0.4147.105), Firefox is ok.
ncp language selection bug

Fix by setting the color to white only if selection is not active.
ncp language selection fixed


nachoparker:
Looks good, thanks for the fix!

nachoparker:
applied. Thanks

PR #1172, opened 2020-08-30
Bump ncp-previewgenerator version to 102.2.1

llucax:
**Disclaimer:** I'm not a Nextcloud developer so I'm not sure if this fix is OK. For more context, I upgraded my ncp instance from NC 18 to NC 19 and now the apps management page says that the app `NCP Preview Generator` at version `102.2.0` is not marked as compatible with NC 19. Checking `/var/www/nextcloud/apps/previewgenerator/appinfo/info.xml`, I see it has the tag `<nextcloud min-version="14" max-version="18" />` and version `<version>102.2.0</version>`.

I checked this repository to see if I was missing a new version of the app, and I see the version in the current master is still 102.2.0 but the NC support was updated to `<nextcloud min-version="14" max-version="19" />`, so I'm assuming that, for an upgrade to happen, the version needs to be actually bumped.

If this is not the correct fix, please let me know how can I get the updated version.

Thanks!

### DIFF
--- a/ncp-previewgenerator/appinfo/info.xml
+++ b/ncp-previewgenerator/appinfo/info.xml
@@ -13,7 +13,7 @@ The first time you activate this app through 'nc-previews-auto', you properly wa
 	</description>
 	<licence>AGPL</licence>
 	<author>Roeland Jago Douma and Ignacio Nunez</author>
-	<version>102.2.0</version>
+	<version>102.2.1</version>
 	<namespace>PreviewGenerator</namespace>
 	<category>multimedia</category>
 	<website>https://nextcloudpi.com</website>


nachoparker:
Thanks for the contribution!

I don't think that's necessary, maybe you didn't have the latest update? (pushed yesterday), also make sure to use ncp-web to upgrade, not the nc updater

llucax:
I used the web updater and it got nothing. Today it seems to be fine, so I guess the version bump is not necessary, as you said. But even so, I think doing a version bump is useful, as it took me a while to figure out why if in the repo version 102.2.0 said that NC version 19 was supported but my nextcloud said it wasn't. If the version didn't match, it would have been much easier to figure out I was missing an update.

ldw12:
@nachoparker 
I also have the same problem. I looked at the updata.sh, it only copies the latest previewgenerator to /var/www/ncp-previewgenerator. the previewgenerator in /var/www/nextcloud/app/ was still the old one. Maybe the update script need to be updated?
I am not a IT guy, just some observations only.

PR #1189, opened 2020-09-28
install smbclient

segator:
smbclient is required to support remote user auth and remote external for smb


nachoparker:
https://github.com/nextcloud/nextcloudpi/blob/master/bin/ncp/CONFIG/nc-nextcloud.sh#L27

nachoparker:
it's already included. Thanks

segator:
Its not working without install smbclient, i tried with latest docker, I
guess user auth is not using php smb client, instead smbclient

Issue #1142, opened 2020-06-13
Ports show closed in nc-info, but are open.


ovpc:
Not really an issue, as ports are in fact open and both NC and NCP web and ncp-config working as expected. Just strange to see ncp reporting closed ports, that are in fact open.
Can someone  look into it and/or explain how this can happen. It is occurring on all my test and production instances. It seems harmless, as it doesn't seem to affect anything, just annoying and confusing. I find the same thing in forum when users include nc-info. Assuming its a bug
Looked at lines 65-75 of bin/ncp-diag, but can not spot it there....

Running nc-info
Gathering information...
NextCloudPi version  v1.26.0
NextCloudPi image    NextCloudPi_03-09-19
distribution         Raspbian GNU/Linux 10 \n \l
automount            no
USB devices          sda 
datadir              /media/xnbox/ncdata
data in SD           no
data filesystem      btrfs
data disk usage      11G/932G
rootfs usage         2,7G/15G
swapfile             /dev/zram3
dbdir                /var/lib/mysql
Nextcloud check      ok
Nextcloud version    18.0.6.0
HTTPD service        up
PHP service          up
MariaDB service      up
Redis service        up
Postfix service      up
internet check       ok
port check 80        closed
port check 443       closed
IP                   192.168.2.26
gateway              192.168.2.254
interface            eth0
certificates         
NAT loopback         no
uptime               9days

dl1dby:
It'not harmless as after updating to NC 18.05 and NCP v1.26 I cannot upload anything. No Webdav, no Caldav. Makes installation pretty unusable. Downloads, Propfind works though.

suxlex:
I have that problem too.

Von: Oliver V <notifications@github.com>
An: nextcloud/nextcloudpi <nextcloudpi@noreply.github.com>
Kopie: Subscribed <subscribed@noreply.github.com>
Empfangen: 13.06.2020 08:38:01
Betreff: [nextcloud/nextcloudpi] Ports show closed in nc-info, but are open. (#1142)

nachoparker:
you are right. I didn't notice this before. The port checker changed its API URL. Easy fix.

Fixed now.

ovpc:
Thank you :-)

FadeFx:
Seems this issue is still not fixed for me in pannel 1.30 and also previous versions. I dont pay much attention to it however, as actually both ports are reachable in reality and letsencrypt renewal also works without issues.

ovpc:
[...]

PR #1194, opened 2020-10-11
- fixed api change for portchecker

dsmic:
This fixes issue #1142


nachoparker:
works great, many thanks!

PR #1213, opened 2020-12-03
removed armhf in favor of base nextcloudpi link

sunjam:
> This code also generates the NextCloudPi docker images for ARM and x86 platforms, and includes an installer for any Debian based system.

Removed armhf from ARM url link.  My understanding from testing is the Docker image supports both armhf and arm64 by simply linking to `ownyourbits/nextcloudpi` when pulling the latest image.


nachoparker:
thanks, how about this https://github.com/nextcloud/nextcloudpi/commit/8743cd1a85ed044b0e7d40cb2a8e26daa4a188e7

PR #1223, opened 2020-12-10
Change time format for backup file name in nc-backup.sh

boofthebear:
Using the Unix time in the backup file name is not user friendly.
Replaced the Unix time (%s) with the absolute time (%H%M%S).


boofthebear:
@nachoparker Any reason why you would stick with the Unix time?
For me there is no reason to have the Unix time in my backup file name.

nextcloud-bkp_20210130_1612043380.tar - using %s
nextcloud-bkp_20210130_165002.tar - using %H%M%S

PR #1245, opened 2021-01-19
nc-autoupdate-ncp: Append to log instead of replace

llucax:
The log is being replaced on each run instead of appended to.


nachoparker:
applied, thanks!

llucax:
Any particular reason why PR aren't being just merged with GitHub?

llucax:
Asking because I can't even see the changes in master (at least yet).

nachoparker:
it's in my local branch. The reason is that I have a script to tag it and add the info from every commit to the changelog

https://github.com/nextcloud/nextcloudpi/blob/master/changelog.md

Issue #1249, opened 2021-01-21
When removing the letsencrypt config it still tries to renew SSL Cert


weidi:
I just placed my NCP instance behind a reverse proxy / traefik that is taking care now of the ssl cert.
Now i keep getting alerts about failed SSL renewal and even after removing the SSL Config from NCP it still seems to try something here.

aphex008:
Same here. Waiting for PR to be accepted.

nachoparker:
PR in `devel` 

thanks

cver:
@weidi would you mind telling me how to setup Nextcloudpi behind Traefik

weidi:
> @weidi would you mind telling me how to setup Nextcloudpi behind Traefik

Hi @cver, this is my docker-compose.yml

````
version: '3'
services:
  nextcloudpi:
    image: ownyourbits/nextcloudpi-x86
    restart: always
    command: "${IP}"
    ports:
#     - "80:80"
#     - "443:443"
     - "4443:4443"
    volumes:
     - ./nextcloud_data:/data
     - /etc/localtime:/etc/localtime:ro
     - ./ncdata:/media/USBdrive
    container_name: nextcloudpi
    networks:
      - proxy
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.nextcloudpi.entrypoints=http"
      - "traefik.http.routers.nextcloudpi.rule=Host(`nextcloud.your-url.com`)"
      - "traefik.http.middlewares.nextcloudpi-https-redirect.redirectscheme.scheme=https"
      - "traefik.http.routers.nextcloudpi.middlewares=nextcloudpi-https-redirect"
      - "traefik.http.routers.nextcloudpi-secure.entrypoints=https"
      - "traefik.http.routers.nextcloudpi-secure.rule=Host(`nextcloud.your-url.com`)"
      - "traefik.http.routers.nextcloudpi-secure.tls=true"
      - "traefik.http.routers.nextcloudpi-secure.tls.certresolver=http"
      - "traefik.http.routers.nextcloudpi-secure.service=nextcloudpi"
      - "traefik.http.services.nextcloudpi.loadbalancer.server.port=80"
      - "traefik.docker.network=proxy"
networks:
  proxy:
    external: true

```

PR #1254, opened 2021-01-31
add remove letsencrypt config

dogwolf:
Based on the issue #1249 , this patch remove files /etc/cron.weekly/letsencrypt-ncp and /etc/letsencrypt/renewal-hooks/deploy/ncp if no domain is set

### DIFF
--- a/bin/ncp/NETWORKING/letsencrypt.sh
+++ b/bin/ncp/NETWORKING/letsencrypt.sh
@@ -46,7 +46,13 @@ configure()
 {
   local DOMAIN_LOWERCASE="${DOMAIN,,}"
 
-  [[ "$DOMAIN" == "" ]] && { echo "empty domain"; return 1; }
+  [[ "$DOMAIN" == "" ]] && {
+     echo "empty domain"; 
+     rm -f /etc/cron.weekly/letsencrypt-ncp
+     rm -f /etc/letsencrypt/renewal-hooks/deploy/ncp  
+     [[ "$DOCKERBUILD" == 1 ]] && update-rc.d letsencrypt disable
+     return 1; 
+  }
 
   # Configure Apache
   grep -q ServerName $vhostcfg && \


nachoparker:
hi, thanks for this.|

I think we would need a couple more things for this to be complete:

- We could add an `ACTIVE` config item (like other NCP apps do) that translates into a checkbox in ncp-web, instead of checking for empty domain.
- Disabling the auto-renewal doesn't stop using the LE certs. We would have to go back to self signed snakeoil certs when disabling. 

In other words, roll back to this

https://github.com/nextcloud/nextcloudpi/blob/master/bin/ncp/CONFIG/nc-nextcloud.sh#L185

dogwolf:
Hi, thanks for the response.
I am sorry but I don't understand this word "Disabling the auto-renewal doesn't stop using the LE certs. We would have to go back to self signed snakeoil certs when disabling."
Do you mean we need to found the SSLCertificateFile parameter on the /etc/apache2/sites-available/nextcloud.conf and /etc/apache2/sites-available/ncp.conf and remove the certification?
Or remove ssl-cert-snakeoil.pem?

Thank you

nachoparker:
We would need to restore these two lines to what they looked like before running letsencrypt for the first time

```
SSLCertificateFile      /etc/ssl/certs/ssl-cert-snakeoil.pem
SSLCertificateKeyFile /etc/ssl/private/ssl-cert-snakeoil.key
```

dogwolf:
> 
> 
> 
>     * We could add an `ACTIVE` config item (like other NCP apps do) that translates into a checkbox in ncp-web, instead of checking for empty domain.
> 

Hi, I tried to edit letsencrypt.cfg, adding

...
 "params": [
    {
      "id": "ACTIVE",
      "name": "Active",
      "value": "no",
      "type": "bool"
    },
    {
      "id": "DOMAIN",
      "name": "Domain",

.....

but, when I open the corresponding web page or the ncp-config page, that param is not showed .. so I suppose it is necessary do something more ..
So Can you tell me which command is necessary in order to show the param?

Thank you

nachoparker:
implemented here https://github.com/nextcloud/nextcloudpi/commit/b8c14093adba8b8c9fd55ee1b691ecb8f98591fb

PR #1287, opened 2021-05-31
[WIP] Fix/diag port check

### LINKED ISSUE #1283: Ports are shown as closed where they are in fact open
Hi, it seems I have a well known problem: NCP shows ports as closed where they are in fact open (as shown by portchecker.co). As a consequence, my installation is unable to refresh the Let´s Encrypt Certificate or do any updates.
My Installation:
<html><body>

NextCloudPi version | v1.34.7
-- | --
NextCloudPi image | NextCloudPi_12-20-19
distribution | Raspbian GNU/Linux 10 \n \l
automount | yes
USB devices | sda sdb sdc
datadir | /media/myCloudDrive/ncdata
data in SD | no
data filesystem | btrfs
data disk usage | 495G/932G
rootfs usage | 6.4G/59G
swapfile | /var/swap
dbdir | /var/lib/mysql/ncdatabase
Nextcloud check | ok
Nextcloud version | 20.0.4.0
HTTPD service | up
PHP service | up
MariaDB service | up
Redis service | up
Postfix service | up
internet check | no
port check 80 | closed
port check 443 | closed

</body>
</html>

theCalcaholic:
Use https for retrieving the public IP from icanhazip.com

Fixes #1283

On the long run it might be worthwhile to check if icanhazip.com is a reliable service or something like DNS services which offer the same functionality (OpenDNS, Google DNS...) should be more stable

(WIP, because I still want to do some more testing)

### DIFF
--- a/bin/ncp-diag
+++ b/bin/ncp-diag
@@ -66,7 +66,18 @@ function is_port_open()
 {
   local port=$1
   local public_ip
-  public_ip="$(curl icanhazip.com 2>/dev/null)" || { echo "closed"; return 1; }
+  public_ip="$(dig +short -4 @resolver1.opendns.com a myip.opendns.com)"
+  [[ "$public_ip" =~ ^([0-9]{1,3}).([0-9]{1,3}).([0-9]{1,3}).([0-9]{1,3})$ ]] \
+    || public_ip="$(dig +short -6 @resolver1.opendns.com aaaa myip.opendns.com)"
+  
+  if [[ "$public_ip" =~ ^";;" ]]
+  then
+    echo "closed"
+    return 1
+  fi
+  
+  public_ip="$(dig +short -4 @resolver1.opendns.com a myip.opendns.com)"
+  [[ 
 
   local tmp_file=$(mktemp)
   local token=$(wget -T2 -t1 -qO- --keep-session-cookies --save-cookies $tmp_file https://portchecker.co | grep -oP "_csrf\" value=\"\K.*\"")
--- a/bin/ncp-diag
+++ b/bin/ncp-diag
@@ -66,18 +66,7 @@ function is_port_open()
 {
   local port=$1
   local public_ip
-  public_ip="$(dig +short -4 @resolver1.opendns.com a myip.opendns.com)"
-  [[ "$public_ip" =~ ^([0-9]{1,3}).([0-9]{1,3}).([0-9]{1,3}).([0-9]{1,3})$ ]] \
-    || public_ip="$(dig +short -6 @resolver1.opendns.com aaaa myip.opendns.com)"
-  
-  if [[ "$public_ip" =~ ^";;" ]]
-  then
-    echo "closed"
-    return 1
-  fi
-  
-  public_ip="$(dig +short -4 @resolver1.opendns.com a myip.opendns.com)"
-  [[ 
+  public_ip="$(curl https://icanhazip.com 2>/dev/null)" || { echo "closed"; return 1; }
 
   local tmp_file=$(mktemp)
   local token=$(wget -T2 -t1 -qO- --keep-session-cookies --save-cookies $tmp_file https://portchecker.co | grep -oP "_csrf\" value=\"\K.*\"")


nachoparker:
sweet! I like https better...

Does this simple change fix the `is_port_open()` function? Can you verify locally with `is_port_open 80` vs `is_port_open 1234`?

Thanks!

dsmic:
the version from fix/diag-port-check does not work for me. I do get a public ip (it is an ipv6 in my case) but if it is send to 

```
  if [[ "${token}" != "" ]]; then
    wget -T2 -t1 -qO- --load-cookies $tmp_file https://portchecker.co --post-data "target_ip=${public_ip}&port=${port}&_csrf=${token::-1}" \
    | grep -q '<span class="green">open</span>' && { echo "open"; return 1; }
  fi

```

it shows closed. I have to remove target_ip=${public_ip} from the call to get it work:
```

  if [[ "${token}" != "" ]]; then
    wget -T2 -t1 -qO- --load-cookies $tmp_file https://portchecker.co --post-data "port=${port}&_csrf=${token::-1}" \
    | grep -q '<span class="green">open</span>' && { echo "open"; return 1; }
  fi

```

nachoparker:
@dsmic the thing is that it always returns ok without the public ip, no matter if the port is open or not, you can compare against 80, assuming you have that open and others that you know are closed

dsmic:
Not on my system:
```

function is_port_open()
{
  local port=$1
  local public_ip
  public_ip="$(curl https://icanhazip.com 2>/dev/null)" || { echo "closed"; return 1; }
    echo $public_ip
  local tmp_file=$(mktemp)
  local token=$(wget -T2 -t1 -qO- --keep-session-cookies --save-cookies $tmp_file https://portchecker.co | grep -oP "_csrf\" value=\"\K.*\"")

  if [[ "${token}" != "" ]]; then
    wget -T2 -t1 -qO- --load-cookies $tmp_file https://portchecker.co --post-data "port=${port}&_csrf=${token::-1}" \
    | grep -q '<span class="green">open</span>' && { echo "open"; return 1; }
  fi
  rm $tmp_file
  echo "closed"
}

echo "port check 1234|$( is_port_open 1234 )"
echo "port check 443|$( is_port_open 443 )"

```

results in:
```
internet check|ok
port check 1234|2003:ce:bf11:bf4b:bd64:c9c2:5b31:154b
closed
port check 443|2003:ce:bf11:bf4b:bd64:c9c2:5b31:154b
open
IP|192.168.2.4
gateway|192.168.2.1
interface|enp3s0
grep: /etc/apache2/sites-available/nextcloud.conf: Datei oder Verzeichnis nicht gefunden
certificates|none
NAT loopback|no

```

dsmic:
OK, after some digging: If I force icanhazip.com (curl -4) to return ipv4 address it works for me with public_ip

```
function is_port_open()
{
  local port=$1
  local public_ip
  public_ip="$(curl -4 https://icanhazip.com 2>/dev/null)" || { echo "closed"; return 1; }
  local tmp_file=$(mktemp)
  local token=$(wget -T2 -t1 -qO- --keep-session-cookies --save-cookies $tmp_file https://portchecker.co | grep -oP "_csrf\" value=\"\K.*\"")

  if [[ "${token}" != "" ]]; then
    wget -T2 -t1 -qO- --load-cookies $tmp_file https://portchecker.co --post-data "target_ip=${public_ip}&port=${port}&_csrf=${token::-1}" \
    | grep -q '<span class="green">open</span>' && { echo "open"; return 1; }
  fi
  rm $tmp_file
  echo "closed"
}

echo "port check 80|$( is_port_open 80 )"
echo "port check 443|$( is_port_open 443 )"


```

This might be the save way to go, if others have problems without public_ip. I think probably portchecker might not support ipv6?!

nachoparker:
Nice! yeah dual IP might have been an issue in your case as well. We are definitely adding this.

> Not on my system

I should have said, that was the case _before_ my last fix, that's why I added `target_ip`. After the fix it shows closed, but I wasn't able to test an open port because I still don't have any.

Thanks for the help

nachoparker:
Good news! I got on the phone with my local ISP and they are going to forward a port temporarily to my router so I can test this as well, even though I don't have IPv6

dsmic:
> Nice! yeah dual IP might have been an issue in your case as well. We are definitely adding this.
> 
> > Not on my system
> 
> I should have said, that was the case _before_ my last fix, that's why I added `target_ip`. After the fix it shows closed, but I wasn't able to test an open port because I still don't have any.
> 
> Thanks for the help

Not on my system with removed public_ip (as in my code snipset). My system shows closed 1234 and open 443 without public_ip . No idea why this is different in your case….

nachoparker:
Interesting, well it was definitely giving me a false positive, so I inspected what the website does internally and they added that parameter at some point, so pretty sure it's good to have that there.

Unless maybe they changed things internally again in the mean time between my test and your test.

nachoparker:
@ovpc also confirmed in private this fixes it. Merging!

Issue #11, opened 2017-05-18
Personal info edit not working


acatanyg:
It is not possible to save the personal infos after edit.

NextCloudPi 11.0.3 & Raspberry Pi 3

nachoparker:
I tested and I could replicate this on NextCloud 12

I found this issue and sure enough, setting the correct time worked for me.

https://github.com/nextcloud/server/issues/4533

Please, check the date on your Raspberry Pi

acatanyg:
Thank you! I will test it next week.

Toni Catany Gelabert

PR #1298, opened 2021-07-07
WIP: NC 21

nachoparker:
TO BE DONE
- [x] test/make it work on docker
- [x] test/make it work on a VM with systemd
- [x] upgrade paths, take into account `notify_push` is only for NC>=21
- [ ] some TODOs


theCalcaholic:
@nachoparker I successfully tested the update from devel to the nc-21 branch on a VM.
How can I best test `notify_push`?
Should I test any other components specifically?

theCalcaholic:
@nachoparker I have successfully tested both the installation on a vm as well as the update path (devel -> nc21).

- How can I best test if `notify_push` is working
- Is there anything else, I should test?

nachoparker:
> @nachoparker I have successfully tested both the installation on a vm as well as the update path (devel -> nc21). All looks well
> 
>     * How can I best test if `notify_push` is working?
> 
>     * Is there anything else, I should test?
> 
> 
> I still have snapshots for every stage of the testing process, so additional tests would not be hard to do

 Many thanks. I think we want to do `ncc notify_push:setup` everytime we change the url, like in letsencrypt and prettyurls.
 
 Not sure what the best way to test this is, maybe just access the `/push` endpoint and see that it is responsive?
 
 In build-VM.sh for instance, we use `nextcloudpi` in `ncc notify_push:setup` during activation, and I think that makes it only accessible locally, so we would have to find a good way of feeding that

theCalcaholic:
Alright, I'll have a closer look at `notify_push` later (or tomorrow).

On that note, are you interested in the test automation setup or the resulting snapshots? I can give you access to my Hetzner cloud project, if you like

theCalcaholic:
@nachoparker Okay, after a few adjustments, `notify_push` is finally working (tested on a VM). I have also added a check for it to system_test.py
The notify_push setup has been moved to a new function in library.sh, `set-nc-domain`. While migrating all apps to said function, I came across a few things, where I wasn't sure if the current implementation is what we really want. I've added Todos in these places, but will also explain them in this PR in form of a review.
Other than that, the next step would be to test the update from devel and the docker containers again :)

bin/ncp/CONFIG/nc-nextcloud.sh
```diff
@@ -203,6 +203,26 @@ EOF
...
+  a2enmod proxy_http
+  a2enmod proxy_wstunnel
+
+  cat > /etc/systemd/system/notify_push.service <<EOF
+[Unit]
+Description = Push daemon for Nextcloud clients
+
+[Service]
+Environment=PORT=7867
+Environment=NEXTCLOUD_URL=https://localhost
```
theCalcaholic, line 217: I hardcoded the url, because there's no harm in using localhost and otherwise, we would have to restart the systemd service every time we changed the NC url

bin/ncp/NETWORKING/dnsmasq.sh
```diff
@@ -73,9 +73,9 @@ EOF
   update-rc.d dnsmasq defaults
   update-rc.d dnsmasq enable
   service dnsmasq restart
-  cd /var/www/nextcloud
-  sudo -u www-data php occ config:system:set trusted_domains 2 --value=$DOMAIN
-  sudo -u www-data php occ config:system:set overwrite.cli.url --value=https://"$DOMAIN"/
+  ncc config:system:set trusted_domains "${TRUSTED_DOMAINS[dnsmasq]}" --value="$DOMAIN"
+  #TODO: Is this really desirable?
+  set-nc-domain "$DOMAIN" --no-trusted-domain
```
theCalcaholic, line 78: Do we really want to overwrite the main domain for nextcloud with the dnsmasq domain? As I understand it, this is the domain of the dns server and there are use cases where the two might differ (e.g. if you only want to use the dns server in the local network and aren't using a global IP/domain)

bin/ncp/NETWORKING/nc-static-IP.sh
```diff
@@ -86,8 +86,9 @@ EOF
     systemctl restart networking
   }
  
-  sudo -u www-data php /var/www/nextcloud/occ config:system:set trusted_domains 1 --value="$IP"
-  sudo -u www-data php /var/www/nextcloud/occ config:system:set overwrite.cli.url --value=https://"$IP"/
+  ncc config:system:set trusted_domains "${TRUSTED_DOMAINS[ip]}" --value="$IP"
+  # TODO: Should we really overwrite potentially configured domains here?
+  set-nc-domain "$IP"
```
theCalcaholic, line 91: Do we really want to overwrite the nc domain with the static IP? Especially since the IP is only reachable locally, that could be problematic. I'd suggest to only add it to trusted_domains but not set overwrite.cli.url

docker/nextcloud/020nextcloud
```diff
@@ -52,15 +52,18 @@ EOF
...
 # Trusted Domain ( as an argument )
 [[ "$2" != "" ]] && \
-  ncc config:system:set trusted_domains 6 --value="$2"
+  ncc config:system:set trusted_domains "${TRUSTED_DOMAINS[docker_overwrite]}" --value="$2"
 
 echo "Provisioning"
 bash /usr/local/bin/ncp-provisioning.sh
 
+echo "Starting notify_push daemon"
+sudo -u www-data /data/nextcloud/apps/notify_push/bin/$(uname -m)/notify_push --allow-self-signed /data/nextcloud/config/config.php &>/dev/null &
```
theCalcaholic, line 65: We could consider removing --allow-self-signed when letsencrypt gets enabled...

etc/library.sh
```diff
@@ -12,6 +12,11 @@ export NCPCFG=${NCPCFG:-/usr/local/etc/ncp.cfg}
 export CFGDIR=/usr/local/etc/ncp-config.d
 export BINDIR=/usr/local/bin/ncp
 
+export TRUSTED_DOMAINS=(
+  [ip]=1 [dnsmasq]=2 [nc_domain]=3 [nextcloudpi-local]=5 [docker_overwrite]=6
+  [nextcloudpi]=7 [nextcloudpi-lan]=8 [public_ip]=11 [letsencrypt_1]=12
+  [letsencrypt_2]=13 [trusted_domain_1]=20 [trusted_domain_1]=21 [trusted_domain_1]=22)
```
theCalcaholic, line 18: This is just so I don't get confused about which slots are used for what and which ones are free :wink:

etc/library.sh
```diff
@@ -101,6 +106,21 @@ function configure_app()
   return $ret
 }
 
+function set-nc-domain() {
+  local DOMAIN="${1?}"
+  URL="https://${DOMAIN%*/}"
+  # trusted_domain no 3 will always contain the overwrite domain
+  [[ "$2" == "--no-trusted-domain" ]] || ncc config:system:set trusted_domains 3 --value="${DOMAIN%*/}"
+  ncc config:system:set overwrite.cli.url --value="${URL}/"
+  if ncc app:enable -q notify_push
```
theCalcaholic, line 115: If we're using the wrong nc version, this if condition evaluates to false

tests/system_tests.py
```diff
@@ -152,6 +153,21 @@ def check_files_dont_exist(files):
     return ret
 
 
+def check_notify_push():
+    "check that notify_push is installed and set up"
+    result = run(pre_cmd + ['ncc', 'notify_push:self-test'], stdout=PIPE, stderr=PIPE)
```
theCalcaholic, line 158: FYI: That's how I tested whether `notify_push` was working

etc/library.sh
theCalcaholic: Btw, I tested that it only fails if the app *can't* be enabled (it succeeds even if it was already enabled)

bin/ncp/CONFIG/nc-nextcloud.sh
nachoparker: are we sure that this doesn't make it only accessible on localhost? I think that `ncc notify_push:setup` complaints about that

etc/library.sh
nachoparker: I've used something like `ncc | grep -q notify_push` in the past, but whatever works is fine

etc/library.sh
nachoparker: I like it

bin/ncp/NETWORKING/nc-static-IP.sh
nachoparker: I think it's that way so we can use pretty urls in this case

bin/ncp/NETWORKING/dnsmasq.sh
nachoparker: yes, we want this for users to be able to use the same domain internally and externally when they don't have hairpin NAT

bin/ncp-update-nc
```diff
@@ -186,6 +186,12 @@ $ncc | grep -q db:add-missing-columns && $ncc db:add-missing-columns -n
 $ncc | grep -q db:add-missing-primary-keys && $ncc db:add-missing-primary-keys -n
 $ncc | grep -q db:convert-filecache-bigint && $ncc db:convert-filecache-bigint -n
 
+# Make sure, the standard for trusted domains is used and notify_push is enabled if possible
+nc_domain="$(ncc config:system:get overwrite.cli.url)"
+nc_domain="${nc_domain#http?://*}"
+nc_domain="${nc_domain%%/*}"
+set-nc-domain "${nc_domain}"
```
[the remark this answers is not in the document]
nachoparker: I think it's a good place for it. We want to make sure it won't fail for older versions without notify_push. Also, we want to install notify_push automatically during the upgrade process

nachoparker:
Another thing I realized is that we need to have two different versions of ncp-previews... one for pre-NC21 and another one post-NC21. This could be a bit thorny to take into account in ncp-update-nc

I am trying to have both copies in the repo, and copy one or another during update depending on current version, but the docker build seems broken ATM (at least in devel).

``
#11 42.70 Running nc-nextcloud
#11 42.75 Downloading Nextcloud 20.0.11...
#11 95.70 Installing  Nextcloud 20.0.11...
#11 97.78 Creating possible missing Directories
#11 97.78 chmod Files and Directories
#11 98.00 chown Directories
#11 98.16 chmod/chown .htaccess
#11 98.17 Starting mariaDB
#11 98.17 mariaDB process not found
#11 98.28 2021-08-07 19:03:00 0 [Note] mysqld (mysqld 10.3.29-MariaDB-0+deb10u1) starting as process 4959 ...
#11 98.68 Setting up database...
#11 98.70 Setting up Apache...
#11 98.71 bash: /usr/local/etc/ncp-templates/nextcloud.conf.sh: No such file or directory
#11 98.71 ERROR: An error occured while generating the nextcloud apache2 config. Attempting safe mode...
#11 98.71 bash: /usr/local/etc/ncp-templates/nextcloud.conf.sh: No such file or directory
#11 98.71 ERROR: Safe mode templating failed as well. Nextcloud will not work.
#11 ERROR: executor failed running [/bin/bash -c set -e; touch /.ncp-image; apt-get update; apt-get install --no-install-recommends -y wget ca-certificates sudo jq; source /usr/local/etc/library.sh; install_app /nc-nextcloud.sh; run_app_unsafe /nc-nextcloud.sh; mv /var/www/nextcloud /data-ro/nextcloud; ln -s /data-ro/nextcloud /var/www/nextcloud; mysqladmin -u root shutdown; apt-get autoremove -y; apt-get clean; find /var/lib/apt/lists -type f | xargs rm; rm -rf /usr/share/man/*; rm -rf /usr/share/doc/*; rm /var/cache/debconf/*-old; rm -f /var/log/alternatives.log /var/log/apt/*; apt-get purge -y wget ca-certificates; rm /nc-nextcloud.sh /usr/local/etc/ncp-config.d/nc-nextcloud.cfg; rm /.ncp-image;]: exit code: 1
```

File: bin/ncp-update-nc (hunk 8 above)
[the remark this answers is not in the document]
Comment (theCalcaholic) [reply bin/ncp-update-nc:193]: It doesn't fail for older versions

File: bin/ncp/CONFIG/nc-nextcloud.sh (hunk 1 above)
Comment (theCalcaholic) [reply bin/ncp/CONFIG/nc-nextcloud.sh:217]: etc)

Comment (theCalcaholic) [reply bin/ncp/CONFIG/nc-nextcloud.sh:217]: That value is only how notify_push connects to Nextcloud (probably for retrieving the list of notifications, etc.)
This configuration is actually right from an example in the notify_push repo

**Comment by theCalcaholic** [comment]
Wait, why is the nextcloud.conf template not copied in docker? I thought I was using the same mechanism that is used for installing/updating ncp configurations...

**Comment by theCalcaholic** [comment]
I definitely need to setup a one-click solution for testing all variants and installation methods for ncp :D 

Currently it's basically at 3 or 4 clicks - but that doesn't seem to be enough to have me test every option before consideringa feature done.

File: bin/ncp-update-nc (hunk 8 above)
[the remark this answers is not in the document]
Comment (theCalcaholic) [reply bin/ncp-update-nc:193]: That's also why I used overwrite.cli.url here instead of trusted_domain 3, because earlier versions didn't use the new consistent assignment of trusted_domain slots

File: bin/ncp/CONFIG/nc-nextcloud.sh (hunk 1 above)
Comment (theCalcaholic) [reply bin/ncp/CONFIG/nc-nextcloud.sh:217]: Just looked it up again: It's [in the Readme](https://github.com/nextcloud/notify_push#troubleshooting) under Troubleshooting:

> If your nextcloud hostname resolves do a dynamic ip you can try setting the NEXTCLOUD_URL to the internal ip of the server.

That's why I think it's both unproblematic and beneficial for us to use localhost (notify_push might not handle dynamic IPs very well)

File: bin/ncp-update-nc (hunk 8 above)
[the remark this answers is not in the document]
Comment (nachoparker) [reply bin/ncp-update-nc:214]: heads up, this does fail when we have `https://localhost`

```
+ ncc config:system:get overwrite.cli.url
+ nc_domain=http://localhost
+ nc_domain=http://localhost
+ nc_domain=http:
+ set-nc-domain http:
+ local domain=http:
+ local url=https://http:
+ [[ '' == \-\-\n\o\-\t\r\u\s\t\e\d\-\d\o\m\a\i\n ]]
+ ncc config:system:set trusted_domains 3 --value=http:
System config value trusted_domains => 3 set to string http:
+ ncc config:system:set overwrite.cli.url --value=https://http:/
System config value overwrite.cli.url set to string https://http:/
```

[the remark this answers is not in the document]
nachoparker: fixed it

nachoparker:
I think that everything works now. Thanks for all the help!!

https://github.com/nextcloud/nextcloudpi/commit/efc55249c20d2f6d26e221a235ab35fc7ec6420b

ovpc:
Thanks you ! Will test it tomorrow, out on a trip today ;-)

PR #1300, opened 2021-07-13
Fix issues with curl installer and integration tests

theCalcaholic:
Fixes some issues with the automated tests as well as the curl installer:

## Issues

1. The automatic nextcloud test failed because nextcloud was complaining about the missing imagick php module. However, it is a deliberate decision to exclude that module in ncp, which renders this test result a false positive.
2. The automatic nextcloud test failed at a second point, because the html id of the login button in Nextcloud changed between versions.
3. During the installation of NCP via the curl installer (install.sh), an error was silently ignored and part of the script was not executed. This was caused by the combination of two bugs:
  1. In the case of errors, the `ncp-update` would abort but still return an exit code of 0 (see [here](https://github.com/nextcloud/nextcloudpi/pull/1300/files#diff-8b7054d41dd575eb79f457c93e62cf10af7be2234d3d48c4c35bcb4f1849aa6dL12))
  2. During installation, `run_update_history.sh` was being executed erroneously, which led to errors and early termination of `ncp.sh` 

## Fixes

- 1 has been fixed [in the automated test itself](https://github.com/nextcloud/nextcloudpi/pull/1300/files#diff-28060f7aeeca88f038c9bd08fd25eb536de66e2dcc71c7219cb2f09552ac26f4R134-R161)
- 2 has also been fixed in the test. [The updated test supports both variants](https://github.com/nextcloud/nextcloudpi/pull/1300/files#diff-28060f7aeeca88f038c9bd08fd25eb536de66e2dcc71c7219cb2f09552ac26f4R122-R128)
- 3.1 has been fixed in [ncp-update](https://github.com/nextcloud/nextcloudpi/pull/1300/files#diff-8b7054d41dd575eb79f457c93e62cf10af7be2234d3d48c4c35bcb4f1849aa6dR12)
- 3.2 has been sort-of fixed by catching any errors in ncp-update and propagating them back to the caller with a custom exit code (see [here](https://github.com/nextcloud/nextcloudpi/pull/1300/files#diff-f643b4c6ad0537aeb471992ff5bb0648d9098b417d284834cf1f570aacef04a8R220-R221) and [here](https://github.com/nextcloud/nextcloudpi/pull/1300/files#diff-f643b4c6ad0537aeb471992ff5bb0648d9098b417d284834cf1f570aacef04a8R304-R307) and [here](https://github.com/nextcloud/nextcloudpi/pull/1300/files#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fR62-R68))

Additionally, the BRANCH variable in the curl installer has been made configurable (and will now propagate through any called scripts that also are branch specific, like `ncp.sh`) to simplify automated testing of arbitrary versions of ncp.

### DIFF
--- a/bin/ncp-update
+++ b/bin/ncp-update
@@ -23,7 +23,7 @@
   ./update.sh || exit $?
 
   cd "$TMPDIR"
-  git describe --always --tags --abbrev=0 > /dev/null 2>&1 | grep -oP "v\d+\.\d+\.\d+" || git fetch --unshallow
+  git describe --always --tags --abbrev=0 2>/dev/null | grep -oP "v\d+\.\d+\.\d+" || git fetch --unshallow --tags
   VER=$( git describe --always --tags --abbrev=0 | grep -oP "v\d+\.\d+\.\d+" )
 
   # check format
--- a/bin/ncp-update
+++ b/bin/ncp-update
@@ -23,6 +23,7 @@
   ./update.sh || exit $?
 
   cd "$TMPDIR"
+  git describe --always --tags --abbrev=0
   git describe --always --tags --abbrev=0 > /dev/null 2>&1 || git fetch --unshallow
   VER=$( git describe --always --tags --abbrev=0 | grep -oP "v\d+\.\d+\.\d+" )
 
--- a/bin/ncp-update
+++ b/bin/ncp-update
@@ -11,7 +11,6 @@
   TMPDIR="$( mktemp -d /tmp/ncp-update.XXXXXX || ( echo "Failed to create temp dir. Exiting" >&2; exit 1 ) )"
   trap 'ret=$?; cd /; rm -rf "${TMPDIR}"; exit $ret' EXIT
 
-  set -x
   echo -e "Downloading updates"
   git clone --depth 20 -b "$BRANCH" -q https://github.com/nextcloud/nextcloudpi.git "$TMPDIR" || {
     echo "No internet connectivity"
@@ -31,8 +30,6 @@
 
   echo "$VER" > /usr/local/etc/ncp-version
   echo "$VER" > /var/run/.ncp-latest-version
-
-  set +x
 
   # write changelog
   git log --graph --oneline --decorate \
--- a/bin/ncp-update
+++ b/bin/ncp-update
@@ -20,7 +20,6 @@
   [[ -f /.ncp-image ]] || cd "$TMPDIR"         # update locally during build
 
   echo -e "Performing updates"
-  [[ -f "/usr/local/etc/ncp-version" ]] || echo "v0.0.0" > /usr/local/etc/ncp-version
   ./update.sh || exit $?
 
   cd "$TMPDIR"
--- a/bin/ncp-update
+++ b/bin/ncp-update
@@ -20,6 +20,7 @@
   [[ -f /.ncp-image ]] || cd "$TMPDIR"         # update locally during build
 
   echo -e "Performing updates"
+  [[ -f "/usr/local/etc/ncp-version" ]] || echo "v0.0.0" > /usr/local/etc/ncp-version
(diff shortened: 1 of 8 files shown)


bin/ncp-update
```diff
@@ -9,7 +9,7 @@
   [[ "$BRANCH" != "master" ]] && echo "INFO: updating to development branch '$BRANCH'"
 
   TMPDIR="$( mktemp -d /tmp/ncp-update.XXXXXX || ( echo "Failed to create temp dir. Exiting" >&2; exit 1 ) )"
-  trap "cd /; rm -rf \"${TMPDIR}\"; exit 0" 0 1 2 3 15
+  trap "cd /; rm -rf \"${TMPDIR}\"" EXIT
```
nachoparker, line 12: I think that this will do

```
  trap 'cd /; rm -rf "${TMPDIR}"; exit $?' EXIT
```

install.sh
```diff
@@ -59,8 +59,13 @@ install_app    lamp.sh
 install_app    bin/ncp/CONFIG/nc-nextcloud.sh
 run_app_unsafe bin/ncp/CONFIG/nc-nextcloud.sh
 systemctl restart mysqld # TODO this shouldn't be necessary, but somehow it's needed in Debian 9.6. Fixme
-install_app    ncp.sh
+ncp_rc=0
```
nachoparker, line 62: what is all this exit code 5 stuff for? if this fails we should bail out I think

tests/nextcloud_tests.py
```diff
@@ -98,19 +119,52 @@ def test_nextcloud(IP, nc_port):
...
+
+            infos = driver.find_elements_by_css_selector("#postsetupchecks > .info > li")
+            if len(infos) != 1:
+                raise ConfigTestFailure("Warnings are shown, but there isn't exactly one info message")
+
+            php_modules = infos[0].find_elements_by_css_selector("li")
+            if len(php_modules) != 1:
+                raise ConfigTestFailure(f"Could not find the list of php modules within the info message "
+                                        f"'{infos[0].text}'")
+            if php_modules[0].text != "imagick":
```
nachoparker, line 155: nice

bin/ncp-update
theCalcaholic: Wouldn't that use the exit code of the rm command?

nachoparker: bruh yes, look again

theCalcaholic: Actually that would do the same as my code snippet, iirc. I looked had something like you are proposing at first, but discovered, that it was unnecessary to explicitly add the exit code inside the trap (because the outer exit code would be passed through by default).

But I'll double check to be sure

install.sh
theCalcaholic: It was my solution for the failing ncp-update, before you told me about `/.ncp-image`.

Should I simply create that file while the installer runs and delete it afterwards (I'll remove the exit 5 stuff once that's working)?

theCalcaholic: Ah sorry, I saw your commit only now. I'll check it first (now I also understand your message on Telegram 🙈)

nachoparker: cool, I am testing some fixes in devel right now

bin/ncp-update
theCalcaholic: I tested it, and it works. The traps passes on the exit code that triggered it by default (if not overwritten)

install.sh
theCalcaholic: I removed the `exit 5` parts :)

tests/nextcloud_tests.py
theCalcaholic: :smiley:

theCalcaholic:
@nachoparker I have now fixed the new issues with the installer. As far as I am concerned, the PR can be merged.

PR #1307, opened 2021-08-04
remove unused USER variable from ncp apps nc-notify-updates and nc-update-nc-apps-auto

rababerladuseladim:
remove unused USER variable from ncp apps nc-notify-updates and nc-update-nc-apps-auto


nachoparker:
good catch. We automated that some time ago and this was left lingering

Issue #1232, opened 2020-12-29
ncp.log every day empty


daschmidt1994:
I almost look daily at my ncp.log for my photo previes. But the ncp.log is every day only a few lines long.

my /etc/logrodate.d/ncp is:

root@nextcloudpi:/# cat etc/logrotate.d/ncp 
/var/log/ncp.log
{
        rotate 4
        size 100M
        missingok
        notifempty
        compress
}

is it possible, that ncp-backup or ncp-rsync-auto the log clear?

**System information**

<details>
<summary>NextCloudPi diagnostics</summary>

```

NextCloudPi version  v1.33.0
NextCloudPi image    NextCloudPi_docker_11-27-20
distribution         Debian GNU/Linux 10 \n \l
automount            no
USB devices          usb1 synoboot 
datadir              /data/nextcloud/data
data in SD           yes
data filesystem      ecryptfs
data disk usage      742G/1.8T
rootfs usage         742G/1.8T
swapfile             /dev/zram3
dbdir                /data/database
Nextcloud check      ok
Nextcloud version    20.0.3.2
HTTPD service        up
PHP service          up
MariaDB service      up
Redis service        up
Postfix service      up
internet check       ok
port check 80        open
port check 443       open
IP                   ***REMOVED SENSITIVE VALUE***
gateway              ***REMOVED SENSITIVE VALUE***
interface            eth0
certificates         ***REMOVED SENSITIVE VALUE***
NAT loopback         no
uptime               3days

```
</details>

<details>
<summary>Nextcloud configuration</summary>

```

{
    "system": {
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": {
          REMOVED
        },
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "dbtype": "mysql",
        "version": "20.0.3.2",
        "overwrite.cli.url": "https:\/\/cloud.daschmidt.at\/",
        "dbname": "***REMOVED SENSITIVE VALUE***",
        "dbhost": "***REMOVED SENSITIVE VALUE***",
        "dbport": "",
        "dbtableprefix": "oc_",
        "mysql.utf8mb4": true,
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "installed": true,
        "instanceid": "***REMOVED SENSITIVE VALUE***",
[...]

PR #1309, opened 2021-08-04
Update nc-backup-auto.sh to append to ncp.log instead of clearing it

schoetju:
append to log instead of clearing it
see https://github.com/nextcloud/nextcloudpi/issues/1232#issuecomment-891524543


schoetju:
First Commit - having trouble with DCO-Check

PR #1342, opened 2021-09-21
Bullseye

nachoparker:
- [x] build ok
- [x] tests pass
- [x] no packages missing -> only php-smbclient -> https://github.com/oerdnj/deb.sury.org/issues/1650
- [x] rpi os support
- [x] armbian support
- [x] upgrade path

### DIFF
--- a/bin/ncp-diag
+++ b/bin/ncp-diag
@@ -56,7 +56,7 @@ fi
 # Services
 echo "HTTPD service|$( pgrep -c apache2 &>/dev/null && echo up || echo down )"
 echo "PHP service|$( pgrep -c php-fpm &>/dev/null && echo up || echo down )"
-echo "MariaDB service|$( pgrep -c mysqld &>/dev/null && echo up || echo down )"
+echo "MariaDB service|$( pgrep -c mariadb &>/dev/null && echo up || echo down )"
 echo "Redis service|$( pgrep -c redis-server &>/dev/null && echo up || echo down )"
 echo "HPB service|$( ncc notify_push:self-test &>/dev/null && echo up || echo down )"
 echo "Postfix service|$( pgrep -fc postfix &>/dev/null && echo up || echo down )"
--- a/bin/ncp-dist-upgrade
+++ b/bin/ncp-dist-upgrade
@@ -42,12 +42,15 @@ apt-get purge -y php${PHPVER} php${PHPVER}-curl php${PHPVER}-gd php${PHPVER}-fpm
 apt-get purge -y php${PHPVER}-mysql
 apt-get purge -y php${PHPVER}-redis
 apt-get purge -y php${PHPVER}-exif
+apt-get purge -y php${PHPVER}-bcmath
+apt-get purge -y php${PHPVER}-gmp
 apt-get purge -y php${PHPVER}-imagick
 set -e
 
 # update sources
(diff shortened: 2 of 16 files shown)


BJKle:
Is there any timeframe, when it will be merged?

BJKle:
@nachoparker  I can see that basically everything is available except the smbclient. On their project site there is no response from the maintainer to you. 
Are there any plans to move forward without that piece of software?

Kirk1984:
It is unlkely that php-smclient will return. it was removed to resolve a bug. Quote:

"Unfortunately, the only solution here is to remove php-smbclient."

Here is the Bugreport:

[https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=952945](url)

Maxxxel:
What you mean by upgrade path? Is this a task of hours, days or weeks? Just wanted to start my own cloud but I'm running bullseye :s

UN-pi:
Since a few days I am running my Pi4 with bullseye NC 22 because there will be a security problem with buster and php 7.3.

grafik
found on [https://www.php.net/supported-versions.php](url)

So php 7.3 will not receive updates after 6.12.21. And php 7.4 will get no security updates 1 year later.
So I decided to upgrade the bullseye php-version to php 8.0  - "assisted" by this sources:

[https://www.c-rieger.de/nextcloud-21-mit-php-8-0-nachruesten/](url) - without special settings.
[https://computingforgeeks.com/how-to-install-php-on-debian-linux/](url)

php8 is shipped with the smbclient and supported until Nov 2023.
I copied the 90-ncp.ini to the fpm\conf.d directory and enabled http2 for apache.

With this configuration I am safe with bullseye and php until Nov. 2023.
Only NCP thinks my bullseye-system is outdated.
Hope that helps.

UN-pi:
I just noticed a few things:

The /usr/local/bin/ncp-diag - file is checking for
buster:
echo "MariaDB service|$( pgrep -c **mysqld** &>/dev/null && echo up || echo down )"

but in bullseye it is running as mariadbd :
echo "MariaDB service|$( pgrep -c **mariadbd** &>/dev/null && echo up || echo down )"

---
I don't know with which configuration php7.4 is enabled by default, but in
/etc/php/8.0/fpm/pool.d/www.conf 

the configuration for pm is set to dynamic **pm = dynamic**
and the other parameters for start_servers etc. are also enabled. Maybe php 7.4 is working in the same way?
So the RAM is used dynamically and normally lower than with static settings.
Enabling the jit-option in php 8.0 for example needs:
; Enabling jit itself by setting th buffer-size
opcache.jit_buffer_size=50M
; configuration for jit s. https://stitcher.io/blog/php-8-jit-setup
opcache.jit=0255

Hope that helps.

BJKle:
> yes, we can do with the binary version, so that's what we did

Sorry, I have to ask again.
Is there a target date or week for a release?

nachoparker:
@UN-pi thanks good catch. thanks

BJKle:
@nachoparker  Busted. After upgrading. NCPI doesn't start anymore. Started fresh (twice) using the current image and upgraded again. Same result. Something is wrong. Can't check since I'm not having a mini hdmi cable.

nachoparker:
please report issues here

https://help.nextcloud.com/t/ncp-upgrade-to-debian-bullseye-thread/128951/3

Maxxxel:
Is this useable now?

Issue #1354, opened 2021-10-02
update is stopped if any running other process name has 'apt' in it


ChristophRg:
Issue is in update.sh line 56
pgrep apt &>/dev/null && { echo "apt is currently running. Try again later";  exit 1; }

Root cause:
pgrep syntax: pgrep 'options' 'pattern' 
pgrep interpretes 'pattern' as regular expression. If a process e.g. called 'capture' is running the update is stopped. 

Fix:
Whole process name to be checked with option -x. Beside apt also apt-get might be checked.
pgrep -x "apt|apt-get" &>/dev/null && { echo "apt is currently running. Try again later";  exit 1; }

nachoparker:
good catch. Care to send a PR?

ChristophRg:
thanks for taking my first PR that fast!

PR #1356, opened 2021-10-05
Update update.sh

ChristophRg:
fix check for running apt during update
(https://github.com/nextcloud/nextcloudpi/issues/1354)

Update script is exiting if any process contains 'apt' in the name. pgrep interpretes <pattern> as regular expression. Whole process name to be checked with option -x.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -52,8 +52,8 @@ samba
 "
 fi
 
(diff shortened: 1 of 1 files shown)


nachoparker:
thanks!

PR #1370, opened 2021-10-28
Clarify documentation of nc-previews-auto command

thomasheller:
At first glance, I thought that "runtime" cloud be used to specify the time when the cron job starts. Had to look at the source code to confirm what the value actually means. Hope this change makes the command easier to understand. :slightly_smiling_face:


nachoparker:
Many thanks!

PR #1372, opened 2021-10-31
Support for encrypted USB drives (LUKS)

thomasheller:
This feature allows to format an external USB drive using `cryptsetup` for LUKS encryption, to store Nextcloud data directory and Nextcloud database securely.

Workflow:
 - Format attached USB drive using nc-luks-format (if it's not already a LUKS-encrypted drive)
 - Open the encrypted container using nc-luks-open (password not stored for security reasons)
 - Enable nc-automount to mount the LUKS partition
 - Move Nextcloud data directory to USB drive using nc-datadir
 - Move Nextcloud database to USB drive using nc-database (optional)
 - Optionally, if you need to detach the USB drive without shutting down the system, use nc-luks-close

Limitations:
 - nc-luks-close cannot unmount the LUKS partition automatically if the Nextcloud database is stored there (You need to move the database back to the original unencrypted location or stop the database server manually)
 - nc-luks-format is not included in the setup wizard

Notes:
 - After rebooting, you need to run nc-luks-open again, because the password is not stored for security reasons. (Until then, you will see the following error message in the Nextcloud frontend: "**Error** Your data directory is invalid Ensure there is a file called `.ocdata` in the root of the data directory." or "Internal Server Error" if you also moved the database to the encrypted drive)

If you have any suggestions for improving this PR, feel free to comment. :slightly_smiling_face:

### DIFF
--- a/bin/ncp/TOOLS/nc-luks-close.sh
+++ b/bin/ncp/TOOLS/nc-luks-close.sh
@@ -9,6 +9,12 @@
 # More at: https://ownyourbits.com
 #
 
+install()
+{
+  apt-get install -y cryptsetup
+  modprobe dm_mod
+}
+
 configure()
 {
   [[ "$DEV" == "" ]] && {
@@ -43,8 +49,6 @@ configure()
 
   echo "successfully unmounted $DEV"
 }
-
-install() { :; }
 
 # License
 #
--- a/bin/ncp/TOOLS/nc-luks-close.sh
+++ b/bin/ncp/TOOLS/nc-luks-close.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+# Unmount and close external USB drive encrypted by LUKS
+#
+# Copyleft 2021 by Thomas Heller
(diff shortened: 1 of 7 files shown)


nachoparker:
Awesome, I really appreciate this, as this is something I have long time wanted to implement.

I also considered LUKS, but then decided for folder only encryption. See the first implementation here https://github.com/nextcloud/nextcloudpi/commit/550ade98de0c4d95741d6437caa916d7e7c21125. I am adding more complete support now (for nc-datadir, and btrfs snapshot support, and eventually also database).

I decided that I don't think it's necessary to encrypt the whole disk, since it brings some complications. When you reboot an nc-encrypt'ed instance, you will be prompted with a password to unlock.

image
image

Given all that, I am not sure if there's still a case for LUKS or if there will be too much overlap and it won't be worth maintaining both approaches.

It is much easier to use the directory encyrption (vs full disk encryption) approach for instances like VMs or docker or anything where the data is in the same logical disk as the OS, since unlocking a root partition with LUKS is awkward and doesn't lend itself to be used in containers.

kojid0:
Using Docker, I prefer the approach of @nachoparker because I see more hurdles having full disk encryption

Issue #1400, opened 2021-12-11
restoring a no-data backup with `nc-restore` + Docker deletes all user data


mackncheesiest:
As I've now learned through experience, restoring an automated backup (`nc-backup-auto`) that doesn't have data included through the nextcloudpi web interface can lead to loss of all user data (:sob:)

Based on my reading of the `nc-restore` script, it looks like it only affects users using the Docker container

First, in the docker container, [this condition passes](https://github.com/nextcloud/nextcloudpi/blob/5a404c00a9e42ea9f8b2e4fa00b50e6182d925c1/bin/ncp/BACKUPS/nc-restore.sh#L27), so we set `NCDIR=/data/nextcloud`

Then, once we check that the backup file is valid, we [remove NCDIR](https://github.com/nextcloud/nextcloudpi/blob/5a404c00a9e42ea9f8b2e4fa00b50e6182d925c1/bin/ncp/BACKUPS/nc-restore.sh#L66) entirely before attempting to do any sort of backup of `$NCDIR/data == /data/nextcloud/data`

At which point we (presumably) end up in the else path of the script [here](https://github.com/nextcloud/nextcloudpi/blob/5a404c00a9e42ea9f8b2e4fa00b50e6182d925c1/bin/ncp/BACKUPS/nc-restore.sh#L132-L146) and create a new data dir if one doesn't exist or scan for existing files (that have all been deleted) if it does.

Based on my reading of the rest of the script and the documentation [here](https://docs.nextcloudpi.com/de/how-to-backup-and-restore-a-nextcloudpi-instance/), this isn't the intent. (`Data-less backups will only restore the Nextcloud configuration files and Nextcloud database. nc-datadir remains untouched.`). Perhaps the `NCDIR` within the docker container shouldn't have all of the `data` located underneath it?

PR #1401, opened 2021-12-12
Prevent backup restoration if datadir within ncdir (#1400)

mackncheesiest:
Follow up to #1400, this is a quick changeset that attempts to fix the bug described there.

Perhaps it would be better in general to try to move the datadir somewhere else if this ever happens. Given that there is already similar logic elsewhere in the script that I'd prefer not to try to coordinate with (since they'd need to know where it gets moved to), this seems like a reasonable way of making script execution quite a bit safer at the cost of being slightly more inconvenient for people who were already planning to replace their datadir.

### DIFF
--- a/bin/ncp/BACKUPS/nc-restore.sh
+++ b/bin/ncp/BACKUPS/nc-restore.sh
@@ -68,7 +68,12 @@ fi
 [[ "$DATADIR" == "" ]] && { echo "Error reading data directory"; exit 1; }
 
 datadir_in_ncdir=$(find "${NCDIR}" -type d -wholename "${DATADIR}")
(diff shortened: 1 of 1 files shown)


mackncheesiest:
I'll give it a shot

mackncheesiest:
Since the existing logic for restoring a with-data backup seemed to be leaving the user's data intact in `$DATADIR-$( date "+%m-%d-%y" )` after restoration, I figured that in that case, this other backup folder can fill that role

However, in a no-data backup restoration I move the folder's contents and remove the folder. Might be nice to have an expression there that doesn't rely on glob expansion since i.e. dotfiles are probably missed here

theCalcaholic:
This should now be fixed (or rather: the default behavior changed in #1601). Sorry for not considering your PR. Frankly, I wasn't aware of it and nacho has left the project as developer

mackncheesiest:
No worries, I'm just happy to see it eventually got fixed! I've migrated away from Nextcloud in general due to a lack of free time to handle issues that came up, but I hope to return to the ecosystem in the future

PR #1403, opened 2021-12-13
Added the firmware v1.32 zip file in the build directory

ZendaiOwl:
Felt it would be alright to have it there for the build process since it is only 3 MB in size and we don't need to query their release page every time the script builds, what do you think?


ZendaiOwl:
I realised a better way of doing this.

I'll close this PR and upload the extracted ISO from the net installer with the UEFI firmware extracted together in one directory. This directory is what's needed to write to an EFI boot partition

Issue #1405, opened 2021-12-17
NCP Docker container doesnt update from v1.45.0


ovpc:
ref : https://help.nextcloud.com/t/ncp-upgrade-from-1-45-0-to-1-46-7-does-not-work/129331/4

> docker exec -it ncp ncp-update
Downloading updates
Performing updates
Nextcloud is already latest version
NextCloudPi updated to version v1.45.0
Error while applying update 1.46.0. Exiting...

Suspect it has to do with fact that there will be fresh images sometime soon (or later), as NCP is moving to Bullseye. It doesn't make sense to dist-upgrade inside a container.

monsieurhannes:
My error message is more like the [initial post](https://help.nextcloud.com/t/ncp-upgrade-from-1-45-0-to-1-46-7-does-not-work/129331?u=hannes) from the forum:
```
root@01bc8618bcad:/# ncp-update
Downloading updates
Performing updates
Error while applying update 1.46.0. Exiting...
root@01bc8618bcad:/#
```

It's basically missing the:
```
Nextcloud is already latest version
NextCloudPi updated to version v1.45.0
```

svenb1234:
So what shall we do? Just wait and see or provide some debug info?

SoongJr:
So far I've only found out that what fails is the execution of this script: https://github.com/nextcloud/nextcloudpi/blob/master/updates/1.46.0.sh

I'm not sure why it fails, running the calls manually does not return an error code, but I did notice a stray quotation mark in there, so maybe that's a clue. Shouldn't affect the execution of _this_ script though, just the execution of the script generated by this code...:
```bash
  cat <<EOF > /etc/cron.daily/refresh_notify_push
#!/usr/bin/env bash
. /usr/local/etc/library.sh
ncc notify_push:self-test || {
  killall notify_push
  sleep 1
  start_notify_push
}"      #<<<<<<< HERE
EOF
```

SoongJr:
Ah shit, it's the check afterwards whether this is NOT in a docker image, or rather, it's the omission of the closing `exit 0`. Previous updates-scripts all ended with an `exit 0` (and they started with a `set -e`), so that's the fault!

SoongJr:
I tested with my fix (by replacing the URL in bin/ncp-update with my fork and telling it to use my feature branch; after the update the script points at the official repo again).
Except for my fork not having had all the tags, and thus my NCP thinking it was not 1.15, the update-script itself worked fine 😅

[...]

PR #1406, opened 2021-12-26
fix update from pre-1.46 to post-1.46 not working in docker

SoongJr:
The script previously used `[[ ! -f /.docker-image ]] && { ... }` as its last statement, which returns non-zero in docker (as it should), but this exit code is then used as the over-all exit code of the update script, thus failing the update whenever used in docker container. Fixed by adding the same `exit 0` that's already there in all other update-scripts. Fixes issue #1405

Also removed a quote that I'm _pretty_ sure shouldn't be there.

### DIFF
--- a/updates/1.46.0.sh
+++ b/updates/1.46.0.sh
@@ -1,3 +1,6 @@
+#!/bin/bash
+
+set -e
 
 # docker images only
 [[ -f /.docker-image ]] && {
@@ -8,7 +11,7 @@ ncc notify_push:self-test || {
   killall notify_push
   sleep 1
   start_notify_push
-}"
+}
 EOF
   chmod +x /etc/cron.daily/refresh_notify_push
 }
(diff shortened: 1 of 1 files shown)


SoongJr:
One minor concern I have is that the addition of `set -e` means that for non-docker images, if any of the following commands fails, the update will now be considered failed where it would previously ignore if any of the first two fails:
```
  systemctl daemon-reload
  systemctl enable refresh_notify_push.{path,service}
  systemctl restart refresh_notify_push.path
```
This _is_ a change in behavior, but I hope a desirable one. Previously only the `systemctl restart` would have caused the update to abort as it's the last statement in the script and its return code becomes the code of the whole script.
Not sure the other two commands can even fail to be honest, just adding this for full disclosure. A test with a non-docker image would definitely be a good idea, but I don't have one ;)

svenb1234:
@nachoparker @theCalcaholic could you please review this commit? Or is this not the correct fix for the issue?

nachoparker:
many thanks for this

PR #1430, opened 2022-03-02
Update build-SD-rpi.sh

ZendaiOwl:

This is a fix for raspbian OS though? The DHCP works just fine one the debian one, when building though you need to copy the /etc/resolv.conf over & /etc/hosts


nachoparker:
oh sorry, we don't want the changes there but in `ncp-dist-upgrade.sh`, my bad

ZendaiOwl:
Ah, alright :+1:

PR #1432, opened 2022-03-02
Update ncp-dist-upgrade

ZendaiOwl:

This is a fix for raspbian OS though, the DHCP works just fine on the debian raspi image, when building though you need to copy the /etc/resolv.conf & /etc/hosts over

Signed-off-by: ZendaiOwl <victorray91@pm.me>


ZendaiOwl:
Ah .. forget the sign off message in the first one ..

nachoparker:
great thanks, yes, you can sign and force-push if you want

ZendaiOwl:
> great thanks, yes, you can sign and force-push if you want

I'll just redo it quickly then, might as well

nachoparker:
I added your signature, thanks!

PR #1433, opened 2022-03-02
Update ncp-dist-upgrade

ZendaiOwl:
This is a fix for raspbian OS, the DHCP works just fine one the raspi debian image, when building though you need to copy the /etc/resolv.conf & /etc/hosts over

Signed-off-by: ZendaiOwl <victorray91@pm.me>


ZendaiOwl:
Ah, I see you already implemented it :+1:

PR #1444, opened 2022-03-24
Add custom metrics exporter

theCalcaholic:
Include custom Prometheus metrics exporter from https://github.com/theCalcaholic/ncp-metrics-exporter. Currently supports only backup monitoring.

**TODO:**

- [x] metrics.sh: Install and manage ncp-metrics-exporter
- [x] Add template ncp-metrics.conf.sh
- [x] nextcloud.conf.sh: Add endpoint /metrics/ncp
- [x] Docker support (including existing metrics)
  - [x] Prevent installation on debian 10 + docker
- ~[ ] Integrate local prometheus and grafana services as an option (feasibility evaluation pending)~

## Testing

*I'm testing the app by installing it into an ncp instance (either via `ncp-update feature/custom-prometheus-exporter` or by directly using the curl installer with this branch) and then activating it via ncp-config. After a backup has been created (e.g. via nc-backup), it should appear in https://<ncp-instance>/metrics/ncp and https://<ncp-instance>/metrics/system should be populated.*

**TODO:**

- [ ] docker 
  - [x] x86_64
    - [x] debian 10
    - [x] debian 11
  - [ ] i386
  - [ ] arm
  - [ ] arm64
- [ ] VM / bare metal
  - [x] x86_64
    - [ ] debian 10
    - [x] debian 11
  - [ ] i386
  - [x] arm64
    - [x] debian 10
    - [x] debian 11
  - [x] arm
    - [x] debian 10
    - [x] debian 11


theCalcaholic:
I have squashed all intermediate commits so the remaining ones are meaningful. As far as I'm concerned, the PR is now ready to be reviewed/merged (I'll probably move the local prometheus/grafana service feature to its own PR, as that's again a lot of work).

theCalcaholic:
I have now squashed all intermediate commits so the remaining ones are meaningful. As far as I'm concerned, the PR is now ready to be reviewed/merged (I'll probably move the local prometheus/grafana service feature to its own PR, as that's again a lot of work).

bin/ncp/BACKUPS/nc-backup-auto.sh
```diff
@@ -7,6 +7,17 @@
...
 #
 
+tmpl_get_destination() {
+  (
+  . /usr/local/etc/library.sh
+  find_app_param nc-backup-auto DESTDIR
+  )
+}
+
+is_active() {
```
nachoparker, line 17: this should not be necessary

nachoparker (commented):
why are we touching backups and nc-datadir here? how is it affected by metrics?

theCalcaholic:
> why are we touching backups and nc-datadir here? how is it affected by metrics?

The new metrics exporter adds backup monitoring. For that to work I need the paths to the backup locations which are stored in a config file at /usr/local/etc/ncp-metrics.cfg. I added a hook to all of the scripts that change backup paths which triggers the regeneration of the config file and a restart of the metrics exporter.

bin/ncp/BACKUPS/nc-backup-auto.sh
theCalcaholic: Alright, I decided against using it to toggle metrics anyways (there's no reason to stop reporting backup freshness, if backups aren't scheduled - quite few opposite, actually). I'll remove it tomorrow. 👍

theCalcaholic:
Good to know. I didn't notice script yet :)

I already tested it in docker and VM, but I want to change it to use 64bit binaries on 64bit architectures. That makes the install script less complicated (no dependencies on different apt packages on different archs) and thus makes ncp instances more consistent and I think that's preferable on the long run.

PR #1457, opened 2022-04-09
ncp-dist-upgrade: Fix invalid target for apt-get on Raspbian

### LINKED ISSUE #1456: upgrade to Bullseye always popping up, but already upgraded
I am using NextcloudPI on a Raspbbery PI for a while now, and I recently upgraded to bullseye, as prompted when login :
`Update to bullseye available. Type 'sudo ncp-dist-upgrade' to upgrade`
The upgrade seemed to have partially worked, I had to install all PHP 7.4 modules manually anyway everything seems to work now.

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Raspbian
Description:	Raspbian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye
```

The issue is I still have the message "Update to bullseye available. Type 'sudo ncp-dist-upgrade' to upgrade" whenever I log in,
 and if I run the ncp-dist-upgrade command, I have a lot of errros while trying to locate php7.3 packages, and above all it ends with 
```
E: The value 'bullseye-security' is invalid for APT::Default-Release as such a release is not available in the sources
Something went wrong. Fix it and try again
```
I am using ncp 1.46.10

theCalcaholic:
Should fix #1456

### DIFF
--- a/bin/ncp-dist-upgrade
+++ b/bin/ncp-dist-upgrade
@@ -69,8 +69,8 @@ apt-get dist-upgrade -y
 
 # install latest PHP version
 release_new=$(jq -r '.release'     < "${new_cfg}")
-# the default repo in bullseye is bullseye-security
-release_new="${release_new}-security"
+# the default repo in bullseye is bullseye-security - except for Raspbian
+[[ -f /usr/bin/raspi-config ]] || release_new="${release_new}-security"
 php_ver_new=$(jq -r '.php_version' < "${new_cfg}")
 
 $APTINSTALL -t ${release_new} php${php_ver_new} php${php_ver_new}-curl php${php_ver_new}-gd php${php_ver_new}-fpm php${php_ver_new}-cli php${php_ver_new}-opcache \


richard2000:
I got an error:
```
ls: cannot access '/etc/php/7.4/fpm/conf.d/*-opcache.ini': No such file or directory
ls: cannot access '/etc/php/7.4/cli/conf.d/*-opcache.ini': No such file or directory
```
In /etc/php/ is only 7.3

richard2000:
I fixed it with this commands:
`sudo apt install php7.4
sudo apt install php7.4-fpm
sudo a2enmod proxy-fcgi setenvif
sudo a2enconf php7.4-fpm
sudo apt install php7.4-intl
sudo apt install php7.4-xml
sudo apt install php7.4-zip
sudo apt install php7.4-mbstring
sudo apt install php7.4-gd
sudo apt install php7.4-curl
`

theCalcaholic:
> I got an error:
> 
> ```
> ls: cannot access '/etc/php/7.4/fpm/conf.d/*-opcache.ini': No such file or directory
> ls: cannot access '/etc/php/7.4/cli/conf.d/*-opcache.ini': No such file or directory
> ```
> 
> In /etc/php/ is only 7.3

That should actually just be a warning and not cause the script to fail, iirc... :thinking:

theCalcaholic:
I merged the PR anyway, since the issue @richard2000 mentioned is caused by the broken state due to the initial failure and not related to the fix itself.

PR #1481, opened 2022-05-20
Fix 32/64 bit architecture detection

DesktopECHO:
To determine if the Debian edition hosting NextCloud is 32 or 64 bit, NCP uses ``uname -m``.  This is mostly fine until you have (for example) a kernel reporting ``aarch64`` that is running Debian ``armhf`` in a container, whereupon the wheels fall off and scripts break.

As opposed to asking the Kernel, it's more reliable to ask the distro if NextCloud is 32 (armv7l, armv8l) or 64 (aarch64) bit to account for all possible hardware setups.

### DIFF
--- a/bin/ncp-update-nc
+++ b/bin/ncp-update-nc
@@ -194,8 +194,8 @@ else
     apachectl -k graceful
     ## make sure the notify_push daemon is runnnig
 
-    arch="$(uname -m)"
-    [[ "${arch}" =~ "armv7" ]] && arch="armv7"
+    arch="$(dpkg --print-architecture)"
+    [[ "${arch}" = "armhf" ]] && arch="armv7"
     cat > /etc/systemd/system/notify_push.service <<EOF
 [Unit]
 Description = Push daemon for Nextcloud clients
--- a/bin/ncp-update-nc
+++ b/bin/ncp-update-nc
@@ -195,7 +195,7 @@ else
     ## make sure the notify_push daemon is runnnig
 
     arch="$(dpkg --print-architecture)"
-    [[ "${arch}" = "armhf" ]] && arch="armv7"
+    [[ "$(uname -m)" =~ "armv7" ]] || [[ "${arch}" == "armhf" ]] && arch="armv7"
     cat > /etc/systemd/system/notify_push.service <<EOF
 [Unit]
 Description = Push daemon for Nextcloud clients
--- a/bin/ncp-update-nc
+++ b/bin/ncp-update-nc
@@ -195,7 +195,7 @@ else
     ## make sure the notify_push daemon is runnnig
 
     arch="$(dpkg --print-architecture)"
-    [[ "$(uname -m)" =~ "armv7" ]] || [[ "${arch}" == "armhf" ]] && arch="armv7"
+    [[ "${arch}" =~ ^(armhf|arm)$ ]] && arch="armv7"
     cat > /etc/systemd/system/notify_push.service <<EOF
 [Unit]
 Description = Push daemon for Nextcloud clients
--- a/bin/ncp/CONFIG/nc-init.sh
+++ b/bin/ncp/CONFIG/nc-init.sh
@@ -152,7 +152,7 @@ EOF
   ncc app:disable updatenotification
 
   # News dropped support for 32-bit -> https://github.com/nextcloud/news/issues/1423
-  if ! [[ "$(uname -m)" =~ "armv7" ]]; then
+  if ! [[ "$(dpkg --print-architecture)" = "armhf" ]]; then
     ncc app:install news
     ncc app:enable  news
   fi
--- a/bin/ncp/CONFIG/nc-init.sh
+++ b/bin/ncp/CONFIG/nc-init.sh
@@ -152,7 +152,7 @@ EOF
   ncc app:disable updatenotification
 
   # News dropped support for 32-bit -> https://github.com/nextcloud/news/issues/1423
-  if ! [[ "$(dpkg --print-architecture)" = "armhf" ]]; then
+  if [[ "$(dpkg --print-architecture)" != "armhf" ]] || [[ ! "$(uname -m)" =~ "armv7" ]]; then
     ncc app:install news
     ncc app:enable  news
   fi
--- a/bin/ncp/CONFIG/nc-init.sh
+++ b/bin/ncp/CONFIG/nc-init.sh
@@ -152,7 +152,7 @@ EOF
   ncc app:disable updatenotification
 
   # News dropped support for 32-bit -> https://github.com/nextcloud/news/issues/1423
-  if [[ "$(dpkg --print-architecture)" != "armhf" ]] || [[ ! "$(uname -m)" =~ "armv7" ]]; then
+  if ! [[ "$(dpkg --print-architecture)" =~ ^(armhf|arm)$ ]]; then
     ncc app:install news
     ncc app:enable  news
   fi
--- a/etc/library.sh
+++ b/etc/library.sh
@@ -165,7 +165,7 @@ function start_notify_push
     if [[ -f /.docker-image ]]; then
       local arch
       arch="$(dpkg --print-architecture)"
-      [[ "${arch}" = "armhf" ]] && arch="armv7"
+      [[ "${arch}" == "armhf" ]] || [[ "$(uname -m)" =~ "armv7" ]] && arch="armv7"
       NEXTCLOUD_URL=https://localhost sudo -E -u www-data /var/www/nextcloud/apps/notify_push/bin/"${arch}"/notify_push --allow-self-signed /var/www/nextcloud/config/config.php &>/dev/null &
     else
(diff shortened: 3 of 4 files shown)


bin/ncp-update-nc
```diff
@@ -194,8 +194,8 @@ else
     apachectl -k graceful
     ## make sure the notify_push daemon is runnnig
 
-    arch="$(uname -m)"
-    [[ "${arch}" =~ "armv7" ]] && arch="armv7"
+    arch="$(dpkg --print-architecture)"
+    [[ "${arch}" = "armhf" ]] && arch="armv7"
```
ZendaiOwl, line 198: It looks good :) 
Tho there should be 2 equal signs inside the double-brackets, I've added a suggestion below which also uses the previous method together with dpkg
```suggestion
    [[ "$(uname -m)" =~ "armv7" ]] || [[ "${arch}" == "armhf" ]] && arch="armv7"
```

bin/ncp/CONFIG/nc-init.sh
```diff
@@ -152,7 +152,7 @@ EOF
   ncc app:disable updatenotification
 
   # News dropped support for 32-bit -> https://github.com/nextcloud/news/issues/1423
-  if ! [[ "$(uname -m)" =~ "armv7" ]]; then
+  if ! [[ "$(dpkg --print-architecture)" = "armhf" ]]; then
```
ZendaiOwl, line 155:
```suggestion
  if [[ "$(dpkg --print-architecture)" != "armhf" ]] || [[ ! "$(uname -m)" =~ "armv7" ]]; then
```
This one should have the exclamation mark inside the double-brackets as we're doing a string comparison, we can also add the other method here with an OR (||) statement and negate the regex (=~) of the 2nd if-statement inside the bracket ( [[ ! ), like I suggested, or outside ( ! [[ ).

etc/library.sh
```diff
@@ -149,8 +149,8 @@ function start_notify_push
     pgrep notify_push &>/dev/null && return
     if [[ -f /.docker-image ]]; then
       local arch
-      arch="$(uname -m)"
-      [[ "${arch}" =~ "armv7" ]] && arch="armv7"
+      arch="$(dpkg --print-architecture)"
+      [[ "${arch}" = "armhf" ]] && arch="armv7"
```
ZendaiOwl, line 153:
```suggestion
      [[ "${arch}" == "armhf" ]] || [[ "$(uname -m)" =~ "armv7" ]] && arch="armv7"
```

ZendaiOwl (changes requested):
It looks good :slightly_smiling_face: 
I have noticed this issue myself when building and experimenting with the docker images, I wrote some suggestions for changes to implent making the if-statements work and also add in the previous check as well just in case, please look it over :pray:

DesktopECHO:
Appreciate the guidance, all the best!
D.

theCalcaholic:
Thanks @DesktopECHO for creating the PR and @ZendaiOwl for reviewing it. I still need to think about the implications of checking with both methods - I'm not 100% sure if that's really what we want yet.

I'll have a more thorough look at it until tomorrow

theCalcaholic:
@DesktopECHO can you please change the target branch for the PR to devel?

ZendaiOwl:
> Thanks @DesktopECHO for creating the PR and @ZendaiOwl for reviewing it. I still need to think about the implications of checking with both methods - I'm not 100% sure if that's really what we want yet.
> 
> I'll have a more thorough look at it until tomorrow

You're welcome :pray:  
I think you're right and we should only use "$(print --dpkg-architecture)" rather than both of them, I thought of keeping it in there as a just in case, you can have a look tomorrow :)

bin/ncp-update-nc
```diff
@@ -194,8 +194,8 @@ else
     apachectl -k graceful
     ## make sure the notify_push daemon is runnnig
 
-    arch="$(uname -m)"
-    [[ "${arch}" =~ "armv7" ]] && arch="armv7"
+    arch="$(dpkg --print-architecture)"
+    [[ "$(uname -m)" =~ "armv7" ]] || [[ "${arch}" == "armhf" ]] && arch="armv7"
```
theCalcaholic, line 198: We should only use one way to detect the architecture
```suggestion
    [[ "${arch}" =~ ^(armhf|arm)$ ]] && arch="armv7"
```

bin/ncp/CONFIG/nc-init.sh
```diff
@@ -152,7 +152,7 @@ EOF
   ncc app:disable updatenotification
 
   # News dropped support for 32-bit -> https://github.com/nextcloud/news/issues/1423
-  if ! [[ "$(uname -m)" =~ "armv7" ]]; then
+  if [[ "$(dpkg --print-architecture)" != "armhf" ]] || [[ ! "$(uname -m)" =~ "armv7" ]]; then
```
theCalcaholic, line 155:
```suggestion
  if ! [[ "$(dpkg --print-architecture)" =~ ^(armhf|arm)$ ]]; then
```

etc/library.sh
```diff
@@ -149,8 +149,8 @@ function start_notify_push
     pgrep notify_push &>/dev/null && return
     if [[ -f /.docker-image ]]; then
       local arch
-      arch="$(uname -m)"
-      [[ "${arch}" =~ "armv7" ]] && arch="armv7"
+      arch="$(dpkg --print-architecture)"
+      [[ "${arch}" == "armhf" ]] || [[ "$(uname -m)" =~ "armv7" ]] && arch="armv7"
```
theCalcaholic, line 153:
```suggestion
      [[ "${arch}" =~ ^(armhf|arm)$ ]] && arch="armv7"
```

updates/1.43.0.sh
```diff
@@ -35,8 +35,8 @@ install_app nc-restore
 [[ ! -f /.docker-image ]] && {
 
   # fix HPB with dynamic public IP
-  arch="$(uname -m)"
-  [[ "${arch}" =~ "armv7" ]] && arch="armv7"
+  arch="$(dpkg --print-architecture)"
+  [[ "${arch}" == "armhf" ]] || [[ "$(uname -m)" =~ "armv7" ]] && arch="armv7"
```
theCalcaholic, line 39:
```suggestion
  [[ "${arch}" =~ ^(armhf|arm)$ ]] && arch="armv7"
```

theCalcaholic (changes requested):
As discussed, I think it is better to only use one way to determine the architecture. I've added corresponding suggestions. :)

updates/1.43.0.sh
ZendaiOwl: Alright :) 

Yes I agree, I left the other one just in case, since it felt like you were sightly hesitating when we were talking in the chat about switching over to `dpkg` 🙏

I'll take a look at the suggestions as soon as I can after work 🥳

theCalcaholic:
(just reopening to retrigger the pipelines, which I fixed in the meantime :P)

theCalcaholic:
@DesktopECHO Could you do a rebase on devel, please?

You can do so  with the following commands:
```sh
git remote add upstream https://github.com/nextcloud/nextcloudpi.git
git fetch upstream
git checkout devel
git reset --hard upstream/devel
git checkout arm32or64
git rebase devel
git push -f
```

theCalcaholic:
After the rebase, I will do some tests and then we should be able to merge this :)

DesktopECHO:
Sorry for the delay, hope this did the trick:

```zero@WinKDE:~/nextcloudpi$ git fetch upstream
zero@WinKDE:~/nextcloudpi$ git checkout arm32or64
Previous HEAD position was b88ce44 nc-datadir.sh: Add success message
Switched to branch 'arm32or64'
Your branch is up to date with 'origin/arm32or64'.
zero@WinKDE:~/nextcloudpi$ git rebase upstream/devel
First, rewinding head to replay your work on top of it...
Applying: Initial Fixes
Applying: Update bin/ncp-update-nc
Applying: Update bin/ncp/CONFIG/nc-init.sh
Applying: Update etc/library.sh
Applying: Update updates/1.43.0.sh
Applying: Update etc/library.sh
Applying: Update updates/1.43.0.sh
Applying: Update bin/ncp-update-nc
Applying: Update bin/ncp/CONFIG/nc-init.sh
zero@WinKDE:~/nextcloudpi$ git push -f origin arm32or64
Enumerating objects: 57, done.
Counting objects: 100% (57/57), done.
Delta compression using up to 36 threads
Compressing objects: 100% (43/43), done.
Writing objects: 100% (47/47), 4.16 KiB | 1.39 MiB/s, done.
Total 47 (delta 41), reused 4 (delta 4)
remote: Resolving deltas: 100% (41/41), completed with 10 local objects.
To https://github.com/DesktopECHO/nextcloudpi.git
 + e92a22b...965ffda arm32or64 -> arm32or64 (forced update)

theCalcaholic:
Seems to work well. Thank you! :)

PR #1503, opened 2022-07-15
let docs link point to actual docs

justDabuK:
**Problem**
The old documentation link doesn't work.

**Solution**
Replaced it with the same link that the [nextcloudpi website](http://www.nextcloudpi.com/) "Documentation Button" points to (https://help.nextcloud.com/t/index-nextcloudpi-documentation/126448#welcome-to-the-nextcloudpi-wiki-1)


schoetju:
I like this (had the same idea)

theCalcaholic:
Thanks for the PR! However, we will have a new documentation landing page launching tomorrow or the day after. After that the URL will work again. Therefore, I'll keep the broken link

PR #1536, opened 2022-08-17
Fix/ncp diag

ZendaiOwl:
I've rewritten `is_port_open` function in `ncp-diag` to now check for open ports on both IPv4 & IPv6, it will also report which type is open, if both are open or closed and it will also give an error message if it encountered an error at the stage of fetching a public IPv4 or IPv6 or if it encountered one when fetching a token for the port check. 

Take a look :pray: 

Signed-off-by: Victor-ray Zendai <victorray91@pm.me>


ZendaiOwl:
Hmm, seems some other commit came with, I'll recreate in another fork .. hang on

PR #1561, opened 2022-09-13
gateway instead of Gateway (as ncp-report looks for this)

schoetju:
ncp-report searches for "gateway" to replace IP with "REMOVED SENSITIVE VALUE"

Signed-off-by: Jürgen <55851807+schoetju@users.noreply.github.com>


ZendaiOwl (approved):
Looks good

PR #1606, opened 2022-10-02
Change backup directory for docker compatibility

MB-Finski:
Improve docker compatibility of "ncp-update-nc.sh" script: Change the backup directory for ./nextcloud/data to reside within the host file system (basedir) so as to avoid moving the data folder back and forth between the docker container and the host file system (results in copying the whole data folder). In situations where the nextcloud data takes up more than the free available space on the host system, this may lead to loss of data.


theCalcaholic:
Thank you for the PR! I'll merge it tomorrow

PR #1607, opened 2022-10-02
Temporarily place backup to BASEDIR

MB-Finski:
Improves docker compatibility. Placing the backup file in /var/www/ will cause a failure in rolling back the update on containerized setups ("Can only restore from ext/btrfs/zfs filesystems"). The host filesystem is most likely compatible so using $BASEDIR as the backup location *should* solve this issue for most users.


theCalcaholic:
Same here - thanks for the PR!

PR #1617, opened 2022-10-08
Update log2ram to latest 1.6.1 release.

lovenemesis:
Update to latest log2ram stable release for the benefit:

- Convert from CRON to systemd.timer, which is supported by NCP base images nowadays.
- rsync by default for better performance.

Personally I have been running 1.6.1 log2ram for a month and it performs just as good as before, if not better.


theCalcaholic:
systemd.timer won't work on docker. Did you test the docker version of NCP?

PR #1626, opened 2022-10-09
Skip $DATADIR when fixing permissions

MB-Finski:
The data directory contains potentially hundreds of thousands files and iterating through all of them thrice will make the upgrade process slow (hours instead of minutes). This arguably results in unnecessary downtime.

This commit proposes skipping fixing permissions/ownership for the data directory since "mv" (should) perserve them when performing a simple upgrade. This would make upgrading considerably faster.

### DIFF
--- a/bin/ncp-update-nc
+++ b/bin/ncp-update-nc
@@ -149,6 +149,13 @@ done
 
 #false # test point
 
+# fix permissions
+####################
+echo "Fix permissions..."
+chown -R www-data:www-data nextcloud
+find nextcloud/ -type d -exec chmod 750 {} \;
+find nextcloud/ -type f -exec chmod 640 {} \;
+
 # copy data if it was at the default location
 ####################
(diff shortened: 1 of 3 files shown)


MB-Finski:
I left this as draft, still, since I'm not 100% sure if there are good reasons why the permissions/ownership should be resetted at each upgrade.

theCalcaholic:
The fix-permissions app is specifically meant to fix permissions for the data directory (e.g. after manually copying files or after restoring a backup) :)

theCalcaholic:
This will soon be relevant, because the data directory will not remain inside /var/www/nextcloud anymore (WIP in https://github.com/nextcloud/nextcloudpi/tree/fix/nc-restore-keep-nc-data)

theCalcaholic:
`ncp-update-nc` was largely rewritten for the upcoming version 24.0.5. Not fixing permissions for the data directory was part of that rewrite :)

PR #1648, opened 2022-10-21
Update update.sh - 32bit architecture check

ZendaiOwl:
Added an architecture check that stops the updater should it detect a 32bit architecture that is trying to update beyond the latest version for Nextcloud 24, which is the final version with 32bit support. 

It will also check what the currently latest version is for NC24 from Nextcloud's changelog website. 

I think this should be enough to prevent 32bit systems from updating beyond a supported release but it would be much appreciated if someone with a 32bit system could actually confirm this :pray: 

I have tested it locally by changing the variables to my "amd64" system and "pretending" it is a 32bit one. 

Signed-off-by: Owl <12261439+ZendaiOwl@users.noreply.github.com>

### DIFF
--- a/bin/ncp-update-nc
+++ b/bin/ncp-update-nc
@@ -68,6 +68,24 @@ then
   exit 1
 fi
 
+# Used to prevent updates beyond Nextcloud v.24, as v.25 does not support 32bit architecture.
+# See System Requirements: https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html
+# See Nextcloud Changelog for current latest versions: https://nextcloud.com/changelog/#latest24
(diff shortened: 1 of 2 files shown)


sunjam:
[25.0.2](https://nextcloud.com/blog/maintenance-releases-25-0-2-24-0-8-and-23-0-12-are-out/) stable now released should support 32-bit hardware until 26.0.0 and my understanding is nc has decided to include a check to prevent all 32-bit systems from upgrading to 26* once it lands... we will see whether that actually happens,  but an fyi.

ZendaiOwl:
Closing this for now, since 32bit was reintroduced for version 25, can look into it later on again

PR #1576, opened 2022-09-27
Fixed: Nextcloudpi icon pointing to ownyoursbits.com

coltkondo:
**Resolved / Related Issues**

- Closes: https://github.com/nextcloud/nextcloudpi/issues/1571

**Comments**

- Updated main logo to re-direct to https://nextcloudpi.com
- Updated right-side header logo to re-direct to https://nextcloudpi.com
- Updated information logo to re-direct to "https://docs.nextcloudpi.com/" (https://github.com/nextcloud/nextcloudpi/wiki is not being used anymore)

PR #1676, opened 2022-12-19
Replace owncloud with nextcloudpi logo

JSchoeck:
Replaced both instances of links to the old ownyourbits.com website with the correct and working link to nextcloudpi.com


theCalcaholic:
Thanks for the PR! However, this was already fixed in https://github.com/nextcloud/nextcloudpi/pull/1576

PR #1677, opened 2022-12-27
Fix/portcheck

ZendaiOwl:
A new way to check if ports are open for external access outside the home network. I changed the layout a little bit in the admin web-interface to reduce the amount of request required from 4 down to 2. Let me know if it looks alright? :pray:

### DIFF
--- a/bin/ncp-diag
+++ b/bin/ncp-diag
@@ -79,6 +79,7 @@ function is_port_open()
   # Checks both port 80 & 443 for IPv4/IPv6 and returns the result or [N/A] [N/A]
   if ! [[ "$TYPE" =~ ^(0|4|6)$ ]]
   then
+    echo "Invalid type: $TYPE" 1>&2
     return 1
   elif [[ "$TYPE" == 0 ]]
   then
--- a/bin/ncp-diag
+++ b/bin/ncp-diag
@@ -91,9 +91,9 @@ function is_port_open()
     return 1
   fi
   # Checks both port 80 & 443 for IPv4
-  test -n "$publicIPv4" && result=($(curl --silent --max-time 4 --ipv4 "$port_url" | jq '."80",."443"' ))
+  test -n "$publicIPv4" && result=($(curl --silent --max-time 4 --ipv4 "$port_url" | jq -r '."80",."443"' ))
   # Check both port 80 & 443 for IPv6
-  test -n "$publicIPv6" && result=($(curl --silent --max-time 4 --ipv6 "$port_url" | jq '."80",."443"' ))
+  test -n "$publicIPv6" && result=($(curl --silent --max-time 4 --ipv6 "$port_url" | jq -r '."80",."443"' ))
   # Prints the result for port 80 & 443 respectively: "open" "open" || "closed" "closed" || "open" "closed" || "closed" "open"
   echo -n "${result[@]}"
 }
--- a/bin/ncp-diag
+++ b/bin/ncp-diag
@@ -79,8 +79,14 @@ function is_port_open()
   # Checks both port 80 & 443 for IPv4/IPv6 and returns the result or [N/A] [N/A]
   if ! [[ "$TYPE" =~ ^(0|4|6)$ ]]
   then
-    echo "Invalid type parameter: $TYPE"
-    return 1
+    if [[ -n "$TYPE" ]]
+    then
+      echo "No type parameter provided"
+      return 1
+    else
+      echo "Invalid type parameter: $TYPE"
+      return 1
+    fi
   elif [[ "$TYPE" == 0 ]]
   then
     echo -e "[N/A]\n[N/A]"
--- a/bin/ncp-diag
+++ b/bin/ncp-diag
@@ -14,8 +14,8 @@
 source /usr/local/etc/library.sh
 
 # Distro, NCP version and tag
-echo "NextCloudPi version|$( cat /usr/local/etc/ncp-version )"
-[[ -f /usr/local/etc/ncp-baseimage ]] && echo "NextCloudPi image|$( cat /usr/local/etc/ncp-baseimage )"
+echo "NextcloudPi version|$( cat /usr/local/etc/ncp-version )"
+[[ -f /usr/local/etc/ncp-baseimage ]] && echo "NextcloudPi image|$( cat /usr/local/etc/ncp-baseimage )"
 echo "OS|$(cat /etc/issue  | sed 's| \\n \\l||'). $(uname -r) ($(uname -m))"
 
 # Data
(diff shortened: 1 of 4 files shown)


ZendaiOwl:
Figured out a way to keep the layout the same, I'll push those changes to this PR

theCalcaholic:
Let's keep this until the NC 25 release next week, so we can get the volunteers to test the current devel version :)

bin/ncp-diag
```diff
@@ -66,78 +66,52 @@ echo "Internet check|$( ping -W 2 -w 1 -q github.com &>/dev/null && echo ok || e
...
+  # The URL leads to an application I've deployed for NCP on https://fly.io using a Docker container I made.
+  # The image for the container is available on Docker Hub (zendai/checkport:sanic) if you wish to deploy one yourself.
+  # The code for the Sanic server and Docker image is available at: https://github.com/ZendaiOwl/Build/tree/master/Docker/Python/Sanic/checkport
+  # I only have a free tier with limited outbound data per month, 100GB p/month.
+  # If we go over 100GB outbound data in a month, I will start being charged for the data going over that limit.
+  # I used a low level Python socket library & fortunately each request only consumes aprox. ~ 60-74 bytes p/second.
+  # Meaning 100GB should be plenty, it should be enough to handle a little less
+  # than 450 request p/second a month, unless my calculations are wrong.
+  # Thank you :pray: from Victor-ray, S. https://github.com/ZendaiOwl
+  local -r PORTURL="https://checkport.zendai.net.eu.org/check"
```
theCalcaholic, line 78: Does this validate the requesting IP? Otherwise your service might be abused by malicious actors to scan for vulnerable servers...

bin/ncp-diag
```diff
@@ -66,78 +66,52 @@ echo "Internet check|$( ping -W 2 -w 1 -q github.com &>/dev/null && echo ok || e
...
+  local TYPE="${1?}" RESULT=("")
+  # Checks both port 80 & 443 for IPv4/IPv6 and returns the result or [N/A] [N/A]
+  test "$TYPE" -eq 4 && RESULT=("$(curl --silent --max-time 4 --ipv4 "$PORTURL" | jq -r '."80",."443"' )")
```
theCalcaholic, line 81: I would prefer to use the bash bracket syntax over the literal test command for consistency with our code base (e.g. `[[ "$TYPE" == 4 ]]` or `[[ "$TYPE" -eq 4 ]]`)

bin/ncp-diag
```diff
@@ -66,78 +66,52 @@ echo "Internet check|$( ping -W 2 -w 1 -q github.com &>/dev/null && echo ok || e
...
+  test "$TYPE" -eq 6 && RESULT=("$(curl --silent --max-time 4 --ipv6 "$PORTURL" | jq -r '."80",."443"' )")
+  test "$TYPE" -eq 0 && RESULT=("[N/A]" "[N/A]")
```
theCalcaholic, line 83: You could reduce code duplication a bit and add parameter validation, e.g.:
```sh
{
# ...
if [[ "$TYPE" =~ ^(0|4|6)$ ]]
then
  echo "Invalid paramter: '$TYPE'" >2
  return 1
elif [[ "$TYPE" == 0 ]]
then
  echo -n "[N/A] [N/A]"
else
  ip_param="--ipv6"
  [[ "$TYPE" == 6 ]] || ip_param="--ipv4"
  curl --silent --max-time 4 "$ip_param" "$PORTURL" | jq -r '."80",."443"'
fi
}
```

bin/ncp-diag
```diff
@@ -66,78 +66,52 @@ echo "Internet check|$( ping -W 2 -w 1 -q github.com &>/dev/null && echo ok || e
...
-}
+test -n "$publicIPv4" && IPv4PORTS=("$(is_port_open 4)") || IPv4PORTS=("$(is_port_open 0)")
+test -n "$publicIPv6" && IPv6PORTS=("$(is_port_open 6)") || IPv6PORTS=("$(is_port_open 0)")
 
-publicIPv4=$(curl -s -m4 -4 "https://icanhazip.com" 2>/dev/null) || unset publicIPv4
-echo "public IPv4|${publicIPv4:-"not found"}"
-publicIPv6=$(curl -s -m4 -6 "https://icanhazip.com" 2>/dev/null) || unset publicIPv6
-echo "public IPv6|${publicIPv6:-"not found"}"
+test "${IPv4PORTS[0]}" == "open" || test "${IPv6PORTS[0]}" == "open" && PORT80="open" || PORT80="closed"
+test "${IPv4PORTS[1]}" == "open" || test "${IPv6PORTS[1]}" == "open" && PORT443="open" || PORT443="closed"
```
theCalcaholic, line 96: Same as above: Please use bash brackets :)

theCalcaholic (changes requested):
Looks great overall! Thanks so much

bin/ncp-diag
ZendaiOwl: The endpoint is a GET only endpoint and checks the ports of the connecting client, it doesn't receive input from anything else, the client ip is fetched from a request header that's handled by the fly.io edge network. 

For validation, it does validate that the client ip is a valid one of ipv4 or ipv6, the ports it checks is hard coded to 80 & 443 together or separately, so there's no way to make it check anything other than the connecting client ip unless spoofing the request. 

However I don't see the point for someone to abuse this maliciously because it's really only useful for someone to check their own port forwarding, cuz if someone is able to spoof their request, they could just check the ports they want directly on whatever host they were scanning with a much larger range instead of going through this endpoint that only checks 80 & 443 

You can see the code for the server here ↓

https://github.com/ZendaiOwl/Build/blob/master/Docker/Python/Sanic/checkport/server.py

bin/ncp-diag
ZendaiOwl: Sure, I can change that 👍

bin/ncp-diag
ZendaiOwl: Oh, yes that looks much better

theCalcaholic: @ZendaiOwl I missed a negation in the first condition. ;) It should be `if ! [[ "$TYPE" =~ ^(0|4|6)$ ]]`

theCalcaholic:
Something weird is going on here... Can you have a look at the docker integration test? Apparently, the NC directory wasn't moved from `/data-ro` to `/data`, which can cause data loss.
I don't see how this could be caused by your PR, though. I have retriggered the docker build, let's see if that resolves it.
Otherwise we should look into this.

theCalcaholic:
Nevermind, I broke it myself :D

theCalcaholic:
Working again :)

ZendaiOwl:
@theCalcaholic Made the changes, using Bash test with double square brackets instead now :)

It also properly handles the return values per new line into an array using `mapfile -t $ARRAY < <(command)`

bin/ncp-diag
```diff
@@ -76,24 +75,57 @@ function is_port_open()
...
-  local TYPE="${1?}" RESULT=("")
+  local TYPE="${1?}" IPType
   # Checks both port 80 & 443 for IPv4/IPv6 and returns the result or [N/A] [N/A]
-  test "$TYPE" -eq 4 && RESULT=("$(curl --silent --max-time 4 --ipv4 "$PORTURL" | jq -r '."80",."443"' )")
-  test "$TYPE" -eq 6 && RESULT=("$(curl --silent --max-time 4 --ipv6 "$PORTURL" | jq -r '."80",."443"' )")
-  test "$TYPE" -eq 0 && RESULT=("[N/A]" "[N/A]")
-  echo -n "${RESULT[@]}"
+  if ! [[ "$TYPE" =~ ^(0|4|6)$ ]]
+  then
+    echo "Invalid type parameter: $TYPE" 2>&1
```
theCalcaholic, line 82: This output redirection has no effect. Do you mean `>&2`, maybe?

theCalcaholic (commented):
looks good :)

bin/ncp-diag
ZendaiOwl: It redirects STDERR to a copy and then into STDOUT :) 

Thought it was something like that you wanted to achieve with this line you wrote as suggestions earlier 

`echo "Invalid paramter: '$TYPE'" >2`

I can change it to `>&2` if you want? 

There isn't really any need for a redirection there at all tho, as the return value will be 0 from echo either way, then the function will return the value 1 by the next line executed

ZendaiOwl: Thinking about it further, there is actually no need for any such echo statements inside there at all, since the result is only dependent on the if statements outside that function, which only sets it to open if it can find a value in the array that equals to "open", otherwise the result will be "closed". So I can just set a `return 1` there

ZendaiOwl:
Alright, now I feel satisfied :) 

It also catches if there wasn't any public IP available now, which should also make it so the suggestions doesn't ask the user to open their ports for Let's Encrypt when they don't have any public IP

bin/ncp-diag
theCalcaholic: I would still keep the echo statement in case the function is reused in the future. Regarding the redirection:

echo prints to STDOUT by default. However, since you interpret the stdout of the function as return value, errors and logs should instead by printed to STDERR, hence the `>&2`.
Otherwise, if a caller doesn't validate the return code correctly, they might try to interprete the error message as port check result

theCalcaholic: output redirection does not affect the return code, btw

theCalcaholic:
> Alright, now I feel satisfied :)
> 
> It also catches if there wasn't any public IP available now, which should also make it so the suggestions doesn't ask the user to open their ports for Let's Encrypt when they don't have any public IP

That's a great addition, thanks!

I've changed the target branch to the (new) release branch for v1.51.0, so we can merge the various changes there

bin/ncp-diag
ZendaiOwl: Oh I see, I'll add an `echo` statement to it again then

ZendaiOwl: Now there is one again

theCalcaholic:
@ZendaiOwl  I cleaned up the commit history and merged the PR :)

For the future, please don't merge base branches into feature branches, rebase the feature branch onto the target branch instead (in this case: `git checkout fix/portcheck; git rebase release/v1.51.0`).

REAPERSbattlecry:
Played a little bit with the devel branch today. 
The new port check works so far. Am I right, that it checks only for an open port not if this port goes to the nextcloud service?

Just to be sure. It really doesn't matter as soon as the port is open,
but if someone reports an issue its even more necessary to ask for more services behind the IP-adress.

ZendaiOwl:
@REAPERSbattlecry 

Yes, it checks if port 80 & 443 is open for external access to a service on the client IP that connects to it by attempting to open a socket connection to the ports 80 & 443 of the same client IP that connected to it with a GET request. (This would be the router's public IP after the network address translation)

This requires a service to be there listening for it to connect to & for that service to be able to accept the incoming connection, so if there is a firewall in the way that blocks the connection it will reported as closed because the service unable to accept the connection - the firewall has blocked it. 

If that service is the Nextcloud service that is set to listen on port 80 & 443, a reverse proxy listening on those ports or another service entirely, it does not detect that, it is not a network mapper like `nmap` in that sense.

The person that sets up the server will have to know what service they have bound to those ports if they are not following the conventional NCP setup. 

If they follow/use the setup of NCP as normal and set by the install script and in the images it will be the Nextcloud service that is listening on port 80 & 443.

PR #1678, opened 2022-12-30
Cleanup

ZendaiOwl:
Cleaned up a little bit in library.sh & changed some NextCloudPi to NextcloudPi in the README.md file. The port check changes are also in here


ZendaiOwl:
Hmm, should I redo this somehow and make it so the other commits are not there?

PR #1679, opened 2022-12-30
࿓❯ Cleaned up a little in library.sh

ZendaiOwl:
- Added tests for the file & directory exports
- Separated ARCH variable declaration, assignmnet & export to not mask the return value
- Single-quoted the keys in the jq command
- Placed brackets around variables where the following word is connected within the double-quotations
- Changed NextCloudPi to NextcloudPi
- Put double-quotes around open variables
- Added '-r' to read to not garble backslashes in input
- Separated local variable declarations & assignment where needed


theCalcaholic:
I'll not merge this just yet, I want to have a look at it first :)
Since it doesn't change functionality, I think it can wait until the next release

ZendaiOwl:
> I'll not merge this just yet, I want to have a look at it first :) Since it doesn't change functionality, I think it can wait until the next release

Yes that's fine :) 

Figured I might as well make some more changes to it so I'll be making a few more pushes to this PR

ZendaiOwl:
Added:

- A function with return codes for checking if a command exists on the system or not, no need to rewrite "command -v $CMD &>/dev/null" every time, can just have one function for that
- A log function for a little more colourized outputs
- A regular install function for package(s) whether you are root or use sudo
- A function that checks if a package is already installed or if it's available in apt

ZendaiOwl:
Adding the information here as well so it's visible

Will be removing some of the latest commits to revert the changes back to the simpler if-statements. After some thought I think it's best I don't introduce unnecessary complexity here at this point, it can be re-worked or refactored when we restructure for the Docker containerization

ZendaiOwl:
There where so many any other commit changes that were present and came along in the git history here of this pull request so I'll just close this PR instead and do the cleanup/refactor of this file during the restructuring for a proper Docker image instead rather than trying to fix the history of this one. I'll make a new Pull Request for the double "/" that is created in one singular change

PR #1682, opened 2023-01-01
࿓❯ Fixed ncp.sh

ZendaiOwl:
Removed the pi user '/sbin/nologin' shell as that user no longer exists by default, the user needs to create the user by themselves using raspberry pi imager and set their own password. Also double-quoted $BRANCH which was left unquoted 

Edit: The user can also be created in `userconf.txt` before booting the image, but better we don't add the user from our side and let the users themselves add one if they want one with RPi imager or the `userconf.txt` file. 

For activating SSH access without having created a user the admin web interface for activating SSH access can be used instead, made it so the user is created if it doesn't exist in `bin/ncp/NETWORKING/SSH.sh`

### DIFF
--- a/ncp.sh
+++ b/ncp.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# NextCloudPi additions to Raspbian
+# NextcloudPi additions to Raspbian
 #
 # Copyleft 2017 by Ignacio Nunez Hernanz <nacho _a_t_ ownyourbits _d_o_t_ com>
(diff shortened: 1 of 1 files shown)


topcaser:
I have the same problem, using the install.sh script on my raspi 64 bit bullseye:
usermod: Benutzer »pi« ist nicht vorhanden.

What can I do to workaround that?

ZendaiOwl:
@topcaser since you already have an existing os up and running you can create the pi user before running the script to allow it to perform the action it's failing to do at the moment

PR #1688, opened 2023-01-07
Improve performance of the "fix permissions" step in ncp-update-nc

powerflo:
The "fix permissions" step is really slow when the nextcloud folder contains a lot of data. In a docker setup the data directory is in nextcloud/data so the folder can be really big. Even on my small test setup the "fix permissions" step took more than 20 minutes. Using "chmod -R ..." reduces the time to around 20 seconds.

Signed-off-by: Florian Wallner <asdf@walura.eu>

### DIFF
--- a/bin/ncp-update-nc
+++ b/bin/ncp-update-nc
@@ -179,8 +179,9 @@ trap rollback INT TERM HUP ERR
 ####################
 echo "Fix permissions..."
 chown -R www-data:www-data nextcloud
-find nextcloud/ -type d -exec chmod 750 {} \;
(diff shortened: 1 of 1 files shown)


schoetju:
I would like to join in saying that improving the performance of “fix permissions” would be really helpful. On my instance the step also takes about 15 minutes (and I get nervous when it doesn't seem to continue during the update). 
As I understand it, the changes suggested by powerflo are functionally identical with significantly better performance. 
@theCalcaholic Is there any concern that the change is patchy or buggy? Can we / community somehow support?

PR #1713, opened 2023-01-22
Fix/create user

ZendaiOwl:
SSH.sh: Rewritten so it now allows for new user creation and doesn't allow for opening up access to the device with the web admin account. 


ZendaiOwl:
Oh dear lord ... Why does GitHub keep having these trailing commits coming in to the PR all the time? So annoying .. 

I'm closing this and making a new pull request from my fork instead

ZendaiOwl:
I don't know why these commits come along with, for some reason `devel` appears to be behind `master` even though they were just merged..

PR #1714, opened 2023-01-22
࿓❯ SSH.sh: Rewrote SSH activation to allow user creation 

ZendaiOwl:
࿓❯ SSH.sh: Rewrote SSH activation so it allows for user creation and disallows the webadmin to be used

### DIFF
--- a/bin/ncp/NETWORKING/SSH.sh
+++ b/bin/ncp/NETWORKING/SSH.sh
@@ -17,7 +17,7 @@ is_active()
 
 configure()
 {
-  [[ $ACTIVE != "yes" ]]  && {
+  [[ "$ACTIVE" != "yes" ]]  && {
     systemctl stop    ssh
     systemctl disable ssh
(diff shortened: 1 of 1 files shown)


ZendaiOwl:
I don't know why these trailing commits come along with, for some reason `devel` appears to be behind `master` even though they were just merged..

theCalcaholic:
@ZendaiOwl you may have to `git rebase devel --signoff`:)

Anyways, the changes look good, but I think, if we allow the creation of new users, we should also disable or delete any old user that was created this way.

Not sure what would be the best way to keep track of that, though...
Maybe add said user to a custom group "ncp-ssh" and look for any existing users in said group?

ZendaiOwl:
Okay :) 

Yes that is a way of keeping track of the created users if it should only be one user created this way,

ZendaiOwl:
Finally managed to squash a couple of the commits

Issue #1624, opened 2022-10-09
default for php max children is too low


theCalcaholic:
Compare #1439 
Currently, the default for the php.ini setting `pm.max_children` is 6, which can easily be exhausted (e.g. by one or two NC desktop clients), causing further requests to timeout.
We should increase this value e.g. to always be >= 16 and <= `SYSTEM_MEMORY / 100MB`

PR #1738, opened 2023-02-26
Improved default for php max children

schoetju:
Like described in issue https://github.com/nextcloud/nextcloudpi/issues/1624
the default for php.ini setting pm.max_children has been improved to be equal total_mem / 100 MB but 16 at minimum.


theCalcaholic:
Thank you!. Looks good after superficial tests. I'll consider it for the next release

Issue #1690, opened 2023-01-08
Let users configure the time for remote synchronizations


theCalcaholic:
**Description**  
`nc-snapshot-sync` and `nc-rsync-auto` allow users regularly copy data to remote endpoints. Since many ncp instances are running behind private internet connections that reconnect to their ISP daily, it would be very useful to be able to adjust the time when the scripts are run, so they aren't interrupted by the router reconnect.

Issue #1623, opened 2022-10-09
Make time for nc-backups-auto configurable


theCalcaholic:
**Description**  
Currently, nc-backup-auto sets up backups to run at 3am. It would be nice to enable users to set the hour for the backups from the web UI.

PR #1762, opened 2023-04-11
Let users configure Hour of Backups / Sync

mayermart:
Add option to configure hour for Auto Backup, Auto RSync and Auto Snapshot via WebUI
Implements #1690 and #1623


theCalcaholic:
Thank you, looks good. I'll add input validation and then merge it into master

PR #1780, opened 2023-05-26
rename ACPU to APCu

szaimen:


ZendaiOwl:
Thank you for your help! @szaimen 

Could you change to `devel` instead of `master` please? 

So it gets merged into `master` from `devel` later :slightly_smiling_face:

szaimen:
> Could you change to `devel` instead of `master` please?

done

ZendaiOwl (approved):
@theCalcaholic For visibility that I'm approving this

PR #1788, opened 2023-06-25
Add DISABLE_FS_CHECK check to nc-database.sh

monkeyclass:
Added check for the env var DISABLE_FS_CHECK when running `nc-database.sh` to change the database directory.


theCalcaholic:
DISABLE_FS_CHECK is meant as a workaround for testing and debugging purposes. However, the next release of NCP adds support for zfs, so the primary use case for this flag should be covered soon

PR #1789, opened 2023-06-26
Add Support for Nextcloud 27

prakharprasad:


theCalcaholic:
Hi, thank you for your PR. However, for the sake of stability, we don't officially support .0 releases of NC. Once NC 27.0.1 is out, we will test and evaluate it and then eventually add support for it.

PR #1860, opened 2023-12-30
Fix bug in nc-previews-auto.sh

m-breitbach:
Fix typo that prevents the feature from being disabled.


theCalcaholic:
Thanks again for the PR. I'll check it and consider it for v1.53.3

theCalcaholic (changes requested):
Hi, this generally looks good, but since the apps name is `nc-previews-auto` not `ncp-previews-auto` I think it would good if the cron file follows this naming scheme. Could you adjust this? Then I'm happy to merge it.

theCalcaholic:
Looks good, thank you!

theCalcaholic (approved):

m-breitbach:
Somehow, the commit didn't make it into the release.

Issue #1818, opened 2023-08-21
Error when upgrading nextcloud: table nextcloud.oc_jobs doesn't exist


janvlug:
I have migrated an old database to nextcloudpi. The old database had no prefixes in the database tables, so I set in the config file dbtableprefix to ''.

But when trying to upgrade nextcloud, I get this:

![image](https://github.com/nextcloud/nextcloudpi/assets/5903208/6a6408f2-7d0d-490c-b57f-7c50df244c21)

table nextcloud.oc_jobs doesn't exist

**System information**

```

NextcloudPi version  v1.52.4
NextcloudPi image    NextCloudPi_RaspberryPi_v1.51.3.img
OS                   Debian GNU/Linux 11. 6.1.21-v8+ (aarch64)
automount            yes
USB devices          sda 
datadir              /media/nextcloudpi_data/ncdata/data
data in SD           no
data filesystem      btrfs
data disk usage      1.3T/5.3T
rootfs usage         4.0G/59G
swapfile             /var/swap
dbdir                /media/USBdrive/ncdatabase
Nextcloud check      ok
Nextcloud version    25.0.9.2
HTTPD service        up
PHP service          up
MariaDB service      up
Redis service        up
HPB service          down
Postfix service      up
Internet check       ok
Public IPv4          ***REMOVED SENSITIVE VALUE***
Public IPv6          ***REMOVED SENSITIVE VALUE***
Port 80              closed
Port 443             closed
IP                   ***REMOVED SENSITIVE VALUE***
Gateway              ***REMOVED SENSITIVE VALUE***
Interface            eth0
Certificates         ***REMOVED SENSITIVE VALUE***
NAT loopback         no
Uptime               3days

```

janvlug:
This error also happens when trying to upgrade from 25.0.9 to 25.0.10:

[ nc-update-nextcloud ] (Tue Aug 22 09:33:53 BST 2023)
Current Nextcloud version 25.0.9.2
Available Nextcloud version 25.0.10
ERROR 1146 (42S02) at line 1: Table 'nextcloud.oc_jobs' doesn't exist

janvlug:
There is a 'jobs' table in my nextcloud database, but no 'oc_jobs' database:

MariaDB [nextcloud]> show tables like 'jobs';
+----------------------------+
| Tables_in_nextcloud (jobs) |
+----------------------------+
| jobs                       |
+----------------------------+
1 row in set (0.002 sec)

[...]

PR #1889, opened 2024-03-31
Fix database prefix

janvlug:
This fixes issue #1818 by getting the database prefix from the configuration file and using it in the SQL command that failed because of a hard coded database prefix.

### DIFF
--- a/bin/ncp-update-nc
+++ b/bin/ncp-update-nc
@@ -74,7 +74,7 @@ fi
 # https://github.com/nextcloud/server/issues/10949
 pgrep -cf cron.php &>/dev/null && { pkill -f cron.php; sleep 3; }
 pgrep -cf cron.php &>/dev/null && { echo "cron.php running. Abort"; exit 1; }
-mysql nextcloud <<<"UPDATE "$DB_FREFIX"jobs SET reserved_at=0;"
+mysql nextcloud <<<"UPDATE ${DB_PREFIX}jobs SET reserved_at=0;"
 
 # cleanup
 ####################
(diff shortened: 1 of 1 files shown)


bin/ncp-update-nc
```diff
@@ -72,7 +74,7 @@ fi
 # https://github.com/nextcloud/server/issues/10949
 pgrep -cf cron.php &>/dev/null && { pkill -f cron.php; sleep 3; }
 pgrep -cf cron.php &>/dev/null && { echo "cron.php running. Abort"; exit 1; }
-mysql nextcloud <<<"UPDATE oc_jobs SET reserved_at=0;"
+mysql nextcloud <<<"UPDATE "$DB_FREFIX"jobs SET reserved_at=0;"
```
theCalcaholic, line 77: Why are you closing and reopening the double quotes here? Did you intend to just quote the variable? Also, there's a typo in the variable's name.

```suggestion
mysql nextcloud <<<"UPDATE ${DB_PREFIX}jobs SET reserved_at=0;"
```

theCalcaholic (changes requested):
Thank you for the PR! Apart from the typo and the quoting it looks good to me and I will merge and test the change once those have been fixed :)

bin/ncp-update-nc
janvlug: IIRC I had initially `mysql nextcloud <<<"UPDATE $DB_PREFIXjobs SET reserved_at=0;"`
which did not work, so I added the closing and reopening quotes, but it could well be that `${DB_PREFIX}` also works.
Sorry for the typo. I know very well that I should always copy and paste :-)

theCalcaholic: Yeah, if you only do `$DB_PREFIXjobs`, it will assume, that's the variable name (including `jobs`). With `${...}` that can be avoided.

theCalcaholic (approved):
Thank you, lgtm now

Issue #1911, opened 2024-05-02
installation via curl failed


ksd73429udmn:
Installed via # curl -sSL https://raw.githubusercontent.com/nextcloud/nextcloudpi/master/install.sh | bash on RP5 on fresh Bookworm. 
 ncp-config returns
cat: /usr/local/etc/ncp-version: No such file or directory
but starts up. 
no options available. 
"NextCloudPi configuration ver. "
Does not show any version.

**System information**

<details>
<summary>NextcloudPi diagnostics</summary>

```

cat: /usr/local/etc/ncp-version: No such file or directory
Nextcloud is not installed - only a limited number of commands are available
grep: /etc/apache2/sites-available/nextcloud.conf: No such file or directory
NextcloudPi version
OS                   Debian GNU/Linux 12. 6.6.28+rpt-rpi-2712 (aarch64)
automount            no
USB devices          none
datadir               (doesn't exist)
rootfs usage         5.2G/28G
swapfile             /var/swap
dbdir                /var/lib/mysql
Nextcloud check      ok
Nextcloud version    28.0.5.1
HTTPD service        up
PHP service          up
MariaDB service      up
Redis service        down
HPB service          down
Postfix service      up
Internet check       ok
Public IPv4          ***REMOVED SENSITIVE VALUE***
Public IPv6          ***REMOVED SENSITIVE VALUE***
Port 80              open
Port 443             open
IP                   ***REMOVED SENSITIVE VALUE***
Gateway              ***REMOVED SENSITIVE VALUE***
Interface            wlan0
Certificates         ***REMOVED SENSITIVE VALUE***
NAT loopback         no
Uptime               38min

```
</details>

<details>
<summary>Nextcloud configuration</summary>

```

Nextcloud is not installed - only a limited number of commands are available


  There are no commands defined in the "config" namespace.



```
</details>

<details>
<summary>HTTPd logs</summary>

```



```
</details>

<details>
<summary>Database logs</summary>

```

tail: cannot open '/var/log/mysql/*.log' for reading: No such file or directory

```
</details>

<details>
<summary>Nextcloud logs</summary>

```

tail: cannot open '/nextcloud.log' for reading: No such file or directory

```
</details>
-->

theCalcaholic:
That likely means that there was a previous error in the installation process. Can you check that or share the full installation log?

[...]

PR #1917, opened 2024-05-03
Fix broken curl installer

NuclearPhoenixx:
Adds `resolvconf` to the `dnsmasq` installation, which should be enough to fix issue #1911 where the installer breaks just after installing `dnsmasq`.


NuclearPhoenixx:
Upon further discussion this does not infact resolve the issue in general, only for Raspberry Pi OS.

PR #2026, opened 2024-12-22
nc-limits.sh: Do not trigger ERR trap due to failed tests

twojstaryzdomu:
The ERR traps introduced in commits e05db188 & ee07434b revealed a bug in `update-nc.sh`, triggered by `bin/ncp-update-nc.d/upgrade-php-bullseye-8.1.sh` or `bin/ncp-update-nc.d/upgrade-php-bookworm-8.3.sh`: Whenever either of the two conditional lists fails in `update-nc.sh`, the `configure` function will return a non-zero status, triggering the ERR trap inherited from the  & executing the rollback from `php_restore`. 

The fix is to ensure the conditional lists do not return a non-zero status.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,17 +27,15 @@ Please reach out in the [Matrix][chat-matrix-wiki] or [Telegram][chat-telegram-w
 
 ---
 
+### Test Status
+
 `master`
 
-[![VM Tests][vm-tests-badge]][vm-tests]
-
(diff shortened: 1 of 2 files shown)


theCalcaholic:
Thank you for the PR! I'll cherry-pick your commit due to the changed base branch and merge that way.

theCalcaholic:
93364e5 has been cherry-picked into devel

PR #2054, opened 2025-04-25
Fix ncpweb performance

sven-ola:
The https://localhost:4443 is very slow on a PI. Main reason is the very slow jq-1.5 on recent Armbians that is executed several times when "source etc/library.sh". One can generally install a faster alternative (see https://github.com/01mf02/jaq) but also this MR implements some fast caching for CPU hogging lines in library.sh

### DIFF
--- a/etc/library.sh
+++ b/etc/library.sh
@@ -14,7 +14,6 @@ export NCDIR=/var/www/nextcloud
 export ncc=/usr/local/bin/ncc
 export NCPCFG=${NCPCFG:-etc/ncp.cfg}
 export ARCH="$(dpkg --print-architecture)"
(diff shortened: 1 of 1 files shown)


bin/ncp-update
```diff
@@ -23,7 +23,7 @@ fi
   trap 'cd /; rm -rf "${TEMPDIR}"' EXIT
 
   echo -e "Downloading updates"
-  git clone --depth 20 -b "$BRANCH" -q https://github.com/nextcloud/nextcloudpi.git "$TEMPDIR" || {
```
TobiGr: Caution: NCP update source was changed

sven-ola: Thanks for the reminder. I was not aware that I was still working on the PullRQ-ed branch. The above change was not meant to be pulled in here. Also, Github complains about signed-offs, thus I will cancel the request.

Issue #2069, opened 2025-09-20
"notify_push" always got resintalled during upgrade


lovenemesis:
This is meant to report [the finding as described in forum](https://help.nextcloud.com/t/upgrade-to-31-0-5-failed-due-to-notify-push/226259?u=lovenemesis).

And a fix to address it.

The fix is tested while upgrading from 31.0.8 to 31.0.9 on a RPi5 image with Armbian.

theCalcaholic:
Fixed in v1.56.0

PR #2070, opened 2025-09-20
#2069: Apply fix for notify_push reinstall during Nextcloud updates

lovenemesis:
Fix to address issue #2069 

Simply adding notify_push add into the copied app list from old instances.

Tested during upgrade from 31.0.8 to 31.0.9 on RPi5 Image with Armbian.


theCalcaholic:
Thank you for the fix! I'll merge it for the next version.

PR #2075, opened 2025-10-16
Fix #2060 invalid ipv6 value

### LINKED ISSUE #2060: Trusted domain entry 11 invalid in ipv6 environment
Trusted Domain entry 11 is invalid when IPv6 is used.

https://github.com/nextcloud/nextcloudpi/blob/2ada653882b625ed91a209969e84cbc9c3d9c0a6/bin/nextcloud-domain.sh#L14
https://github.com/nextcloud/nextcloudpi/blob/2ada653882b625ed91a209969e84cbc9c3d9c0a6/bin/nextcloud-domain.sh#L16 

Option -m4 seems to be undefined, hence  `curl -m4 icanhazip.com` always returns IPv6.

Solution
1) if ipv6 is returned then the value should be enclosed in brackets „[]“ when set (line 16)
2) run `curl -4 icanhazip.com ` in order to always return ipv4 (line 14)

lukaszgoworko:
this should fix the error when restarting the raspi and renewing the trusted domains with ipv6 values where the ipv6 value is missing the enclosing brackets.

fixes: #2060


bin/nextcloud-domain.sh
```diff
@@ -11,9 +11,12 @@ done
 # wicd service finishes before completing DHCP
 while :; do
   local_ip="$(get_ip)"
-  pub_ip="$(curl -m4 icanhazip.com 2>/dev/null)"
 
-  [[ "$pub_ip"   != "" ]] && ncc config:system:set trusted_domains 11 --value="$pub_ip"
+  pub_ipv4="$(curl -4 -m4 icanhazip.com 2>/dev/null)"
+  pub_ipv6="$(curl -6 -m4 icanhazip.com 2>/dev/null)"
+  [[ "$pub_ipv4"   != "" ]] && ncc config:system:set trusted_domains 11 --value="$pub_ipv4"
+  [[ "$pub_ipv6"   != "" ]] && ncc config:system:set trusted_domains 12 --value="[$pub_ipv6]"
```
theCalcaholic: This syntax is preferrable:
```suggestion
  [[ -z "$pub_ipv6" ]] || ncc config:system:set trusted_domains 12 --value="[$pub_ipv6]"
```

theCalcaholic (approved):
LGTM, I'll add the two minor syntax improvements.
Thanks for the fix!

PR #2090, opened 2026-01-28
fix: replace legacy /etc/issue with /etc/os-release

CrazyWolf13:
After upgrading to Debian 13, NextcloudPi still reported Debian 12.

This was traced to ncp-diag, which currently reads /etc/issue to determine the OS version. /etc/issue is a legacy login banner and is not updated during distribution upgrades.

This PR switches OS detection to /etc/os-release, which is the standard and is automatically updated on dist-upgrades.

```bash
# before
OS | Debian GNU/Linux 12. 6.17.4-2-pve (x86_64)
# after
OS | Debian GNU/Linux 13 (trixie). 6.17.4-2-pve (x86_64)
```


theCalcaholic:
Great catch, thank you

PR #2105, opened 2026-04-05
Fix backup cache handling when JSON decode returns null

Dominik0101:
The WebUI crashes when the cache file contains invalid or empty JSON.

This happens because json_decode($cache_str, true) may return null.
The previous implementation used `or [ ]`, which does not assign the fallback array as intended due to PHP operator precedence.

This patch ensures that $cache is always initialized as an array:
- If json_decode returns a valid array, it is used
- Otherwise, $cache falls back to an empty array

This prevents runtime errors when accessing the cache.

### DIFF
--- a/ncp-web/backups.php
+++ b/ncp-web/backups.php
@@ -55,7 +55,10 @@ HTML;
     $cache_str = file_get_contents($cache_file)
       or exit("error opening ${cache_file}");
 
-    $cache = json_decode($cache_str, true) or [];
+    $cache = json_decode($cache_str, true);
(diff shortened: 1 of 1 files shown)


theCalcaholic:
Thanks so much, lgtm

theCalcaholic:
The pipelines don't seem to be working on a different repo, so I'll merge this into a temporary branch for testing. :)
It's a small enough change, so I don't expect anything to come up

theCalcaholic:
Continued in #2113

Issue #2082, opened 2025-11-12
Debian 13 Support


sunjam:
For tracking support of Debian 13, which will close this issue.

Debian 12
Standard support EOL: June 30, 2026
LTS (Long Term Support) EOL: June 30, 2028

MagelanM:
I am working on a PR for this. 
I’d be grateful if anyone could if somebody will support me.
[https://github.com/MagelanM/nextcloudpi/tree/feature/Debian-13-Support-2082](url)

MagelanM:
PR for this issue is ready: #2122

Krawei:
@theCalcaholic seems that this needs topic here needs your attention...

sunjam:
Adjusting that link to show anyone browsing here that it was merged.  Will close this issue as completed.  
- https://github.com/nextcloud/nextcloudpi/pull/2122

theCalcaholic:
@sunjam Thank you!

PR #2122, opened 2026-05-31
Feature/debian 13 support / Issue #2082

MagelanM:
# Description
This feature should update Debian 12 (bookworm) to Debian 13 (trixie).
The security support for Debian 12 end on 10 June 2026. With Debian LTS supports ends 20 August 2028.

# Changes
- switch to trixie dist repos
- suppresses stderr-Output if PHP is not yet installed (new system)
- adoption for maria db install command
- OSCP stapling
    - Disable OCSP stapling for self-signed certs - fixes OpenSSL 3.5 incompatibility on trixie
    - Enable OCSP stapling with valid Let's Encrypt certificate
- ssh: reload via socket on trixie
- dist-upgrade: remove switch to systemd-resolved for lxc enviroments, it breaks DNS

# Not changed
- PHP version 8.3. is unchanged

# Tests
### Install on plain Debian 13 LXC
- ncp-config → UPDATES → nc-update-nextcloud 33.0.2 -> 33.0.3
- ncp-config → BACKUP → nc-backup
- ncp-config → DATABASE → nc-database → move db dir
- ncp-config → UPDATES → unattended-upgrades

### Dist upgrade on NCP 1.57.1, NC 33.0.2
- ncp-config → UPDATES → nc-update-nextcloud 33.0.2 -> 33.0.3
- ncp-config → BACKUP → nc-backup
- ncp-config → UPDATES → nc-update-nc-apps
- ncp-config → UPDATES → unattended-upgrades
- Check for maria db warn/errors (journalctl -u mariadb --no-pager | grep -i 'warn\|error')

### DIFF
--- a/bin/ncp-update-nc.d/update-nc.sh
+++ b/bin/ncp-update-nc.d/update-nc.sh
@@ -260,6 +260,11 @@ then
 
   # Reload library.sh to reset PHPVER
   source /usr/local/etc/library.sh
+elif is_more_recent_than "8.3.0" "${PHPVER}.0" && [[ "$DEBIAN_VERSION" -ge 13 ]]
+then
+  /usr/local/bin/ncp-update-nc.d/upgrade-php-trixie-8.3.sh
+  # Reload library.sh to reset PHPVER
+  source /usr/local/etc/library.sh
 fi
 
 # refresh completions
--- a/bin/ncp-update-nc.d/upgrade-php-trixie-8.3.sh
+++ b/bin/ncp-update-nc.d/upgrade-php-trixie-8.3.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+source /usr/local/etc/library.sh
+
+echo "Refreshing PHP repository for trixie..."
+export DEBIAN_FRONTEND=noninteractive
+
+# Refresh sury repo for trixie
+wget -O /etc/apt/trusted.gpg.d/php.gpg https://packages.sury.org/php/apt.gpg
+echo "deb https://packages.sury.org/php/ ${RELEASE%-security} main" > /etc/apt/sources.list.d/php.list
+apt-get update
+
+echo "PHP ${PHPVER} repository updated for trixie."
--- a/bin/ncp/CONFIG/nc-datadir.sh
+++ b/bin/ncp/CONFIG/nc-datadir.sh
@@ -8,6 +8,10 @@
 # More at https://ownyourbits.com/2017/03/13/nextcloudpi-gets-nextcloudpi-config/
 #
 
+log_step() {
+  echo "[$(date -Is)] $*"
+}
+
 is_active()
 {
   local SRCDIR
@@ -112,6 +116,10 @@ configure()
     btrfs subvolume create "${BASEDIR}"
   }
 
+  log_step "setting nc config: datadirectory"
+  # first set in config the new value, because occ checks if the current datadir exists
+  ncc config:system:set datadirectory --value="${DATADIR}"
+
   # use encryption, if selected
   if is_active_app nc-encrypt; then
     # if we have encryption AND BTRFS, then store ncdata_enc in the subvolume
@@ -123,12 +131,7 @@ configure()
   fi
   chown www-data: "${DATADIR}"
 
-  # datadir
-  ncc config:system:set datadirectory  --value="${DATADIR}" \
-  || sed -i "s|'datadirectory' =>.*|'datadirectory' => '${DATADIR}',|" "${NCDIR?}"/config/config.php
-
-  ncc config:system:set logfile --value="${DATADIR}/nextcloud.log" \
-  || sed -i "s|'logfile' =>.*|'logfile' => '${DATADIR}/nextcloud.log',|" "${NCDIR?}"/config/config.php
+  ncc config:system:set logfile --value="${DATADIR}/nextcloud.log"
   set_ncpcfg datadir "${DATADIR}"
 
   # tmp upload dir
--- a/bin/ncp/CONFIG/nc-datadir.sh
+++ b/bin/ncp/CONFIG/nc-datadir.sh
@@ -8,10 +8,6 @@
 # More at https://ownyourbits.com/2017/03/13/nextcloudpi-gets-nextcloudpi-config/
 #
 
-log_step() {
-  echo "[$(date -Is)] $*"
-}
-
 is_active()
 {
   local SRCDIR
@@ -116,7 +112,6 @@ configure()
     btrfs subvolume create "${BASEDIR}"
   }
 
-  log_step "setting nc config: datadirectory"
   # first set in config the new value, because occ checks if the current datadir exists
   ncc config:system:set datadirectory --value="${DATADIR}"
 
--- a/bin/ncp/CONFIG/nc-init.sh
+++ b/bin/ncp/CONFIG/nc-init.sh
@@ -180,7 +180,6 @@ EOF
   if ! is_more_recent_than "21.0.0" "${ncver}"; then
     ncc app:install notify_push
     ncc app:enable  notify_push
-    test -f /.ncp-image || start_notify_push # don't start during build
   fi
 
   # previews
@@ -215,6 +214,9 @@ EOF
   test -f /usr/local/bin/nextcloud-domain.sh && {
     test -f /.ncp-image || bash /usr/local/bin/nextcloud-domain.sh
   }
+
+  # start notify_push after NC is fully configured
+  test -f /.ncp-image || start_notify_push # don't start during build
 
   # dettach mysql during the build
   if [[ "${db_pid}" != "" ]]; then
--- a/bin/ncp/NETWORKING/freeDNS.sh
+++ b/bin/ncp/NETWORKING/freeDNS.sh
@@ -7,10 +7,25 @@
 #
 
 
-install()
-{
-  apt-get update
-  apt-get install --no-install-recommends -y dnsutils
+wait_for_dpkg() {
+  local tries=0
+  while fuser /var/lib/dpkg/lock-frontend /var/lib/dpkg/lock >/dev/null 2>&1; do
+    echo "dpkg locked, waiting..."
+    fuser -v /var/lib/dpkg/lock-frontend /var/lib/dpkg/lock 2>/dev/null || true
+    sleep 2
+    tries=$((tries + 1))
+    if [[ $tries -ge 150 ]]; then
+      echo "dpkg lock timeout"
+      return 1
+    fi
+  done
+}
+
+install() {
+  wait_for_dpkg || return 1
+  apt-get update || return 1
+  wait_for_dpkg || return 1
+  DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y dnsutils || return 1
 }
 
 configure() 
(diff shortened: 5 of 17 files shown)


bin/ncp-update-nc.d/update-nc.sh
```diff
@@ -258,6 +258,11 @@ elif ! is_more_recent_than "29.0.0" "${NCVER}" && is_more_recent_than "8.3.0" "$
 then
   /usr/local/bin/ncp-update-nc.d/upgrade-php-bookworm-8.3.sh
 
+  # Reload library.sh to reset PHPVER
+  source /usr/local/etc/library.sh
+elif is_more_recent_than "8.3.0" "${PHPVER}.0" && [[ "$DEBIAN_VERSION" -ge 13 ]]
```
theCalcaholic, line 263: I would rather block the dist-upgrade until PHP is upgraded to 8.3+
That way the PHP upgrade always has the same environment and doesn't need to be implemented and tested twice.

bin/ncp-update-nc.d/upgrade-php-trixie-8.3.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
theCalcaholic, line 1: see previous comment - I'd rather avoid doing the upgrade to PHP 8.3 on trixie at all.

bin/ncp/CONFIG/nc-datadir.sh
```diff
@@ -112,6 +112,9 @@ configure()
     btrfs subvolume create "${BASEDIR}"
   }
 
+  # first set in config the new value, because occ checks if the current datadir exists
+  ncc config:system:set datadirectory --value="${DATADIR}"
```
theCalcaholic, line 116:
```suggestion
  ncc config:system:set datadirectory --value="${DATADIR}" \
  || sed -i "s|'datadirectory' =>.*|'datadirectory' => '${DATADIR}',|" "${NCDIR?}"/config/config.php
```

let's keep the fallback in case occ doesn't work (not sure if it's needed, but I'd rather keep it than have it fall apart at some point).

bin/ncp/CONFIG/nc-datadir.sh
```diff
@@ -123,12 +126,7 @@ configure()
   fi
   chown www-data: "${DATADIR}"
 
-  # datadir
-  ncc config:system:set datadirectory  --value="${DATADIR}" \
-  || sed -i "s|'datadirectory' =>.*|'datadirectory' => '${DATADIR}',|" "${NCDIR?}"/config/config.php
-
-  ncc config:system:set logfile --value="${DATADIR}/nextcloud.log" \
-  || sed -i "s|'logfile' =>.*|'logfile' => '${DATADIR}/nextcloud.log',|" "${NCDIR?}"/config/config.php
+  ncc config:system:set logfile --value="${DATADIR}/nextcloud.log"
```
theCalcaholic, line 129:
```suggestion
  ncc config:system:set logfile --value="${DATADIR}/nextcloud.log" \
  || sed -i "s|'logfile' =>.*|'logfile' => '${DATADIR}/nextcloud.log',|" "${NCDIR?}"/config/config.php
```

see above

bin/ncp/CONFIG/nc-init.sh
```diff
@@ -216,6 +215,9 @@ EOF
     test -f /.ncp-image || bash /usr/local/bin/nextcloud-domain.sh
   }
 
+  # start notify_push after NC is fully configured
+  test -f /.ncp-image || start_notify_push # don't start during build
```
theCalcaholic, line 219: doesn't look like a problem, however, I'm wondering: did you need this to fix anything?

bin/ncp/NETWORKING/freeDNS.sh
```diff
@@ -7,10 +7,25 @@
 #
 
 
-install()
-{
-  apt-get update
-  apt-get install --no-install-recommends -y dnsutils
+wait_for_dpkg() {
```
theCalcaholic, line 10: can you explain the changes to this file? They seem unrelated to the dist-upgrade ...

bin/ncp/NETWORKING/letsencrypt.sh
```diff
@@ -62,6 +62,8 @@ configure()
     local key_path="$(grep SSLCertificateKeyFile "${nc_vhostcfg}" | awk '{ print $2 }')"
     sed -i "s|SSLCertificateFile.*|SSLCertificateFile ${cert_path}|"      "${ncp_vhostcfg}"
     sed -i "s|SSLCertificateKeyFile.*|SSLCertificateKeyFile ${key_path}|" "${ncp_vhostcfg}"
+    # Enable OCSP stapling with valid Let's Encrypt certificate
+    sed -i 's/SSLUseStapling.*$/SSLUseStapling          on/' /etc/apache2/conf-available/http2.conf
```
theCalcaholic, line 66:
```suggestion
    sed -i 's/SSLUseStapling.*$/SSLUseStapling          off/' /etc/apache2/conf-available/http2.conf
```

If I understand correctly, what you intend here, this should be `off` in this case, right?

etc/ncp-templates/mysql/91-ncp.cnf.sh
```diff
@@ -14,9 +14,6 @@ fi
 cat <<EOF
 [mysqld]
 transaction_isolation = READ-COMMITTED
-innodb_large_prefix=true
```
theCalcaholic: I assume, these removed options are deprecated? Can you explain the change to me briefly? :)

etc/ncp-templates/systemd/php-fpm.service.d.ncp.conf.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
theCalcaholic, line 1: Thanks so much, this would probably have taken me a good amount of time to figure out (especially for the LXC case).

update.sh
```diff
@@ -172,6 +172,11 @@ if is_docker; then
   cp build/docker/{lamp/010lamp,nextcloud/020nextcloud,nextcloudpi/000ncp} /etc/services-enabled.d
 fi
 
+# PHP-FPM systemd drop-in: allow write access to NCP paths blocked by ProtectSystem=full
```
theCalcaholic, line 175: Correct me if I'm wrong, but I believe, that this is only needed for systems that are updated from bookworm, right?
If so, can you move it to `bin/ncp-dist-upgrade.d/debian-12.sh`?

theCalcaholic (changes requested):
Thank you so much! Overall this looks great and I appreciate the work you have put into this.

I have left a number of review comments with a few minor suggestions and a few questions. Please answer/react to those before we can merge this.

bin/ncp-update-nc.d/update-nc.sh
MagelanM: You're right. The PHP version check should be removed from the condition.

bin/ncp-update-nc.d/upgrade-php-trixie-8.3.sh
MagelanM: Yes, that ties in with the previous comment. I thought I had to do it the same way as the others and wanted to change as little as possible.

bin/ncp/CONFIG/nc-datadir.sh
MagelanM: That's a tricky issue.
In principle, I understand that it's good to have a fallback. But then it would have to be implemented consistently.
From my perspective, any place where manual changes are made to the configuration is a potential source of error. But that’s a recurring issue throughout NCP.
First, I’d naively ask: Can’t we ensure that the OCC works? If we can’t do that, then I think we have a completely different problem. But I can’t see the whole picture...
If we need the fallback, I'm not sure where that's actually implemented. In other words, where occ is started up and the modified files are actually read in.
A few steps later, we have calls where occ is active again but isn't aware of the changes in the files. I encountered this error during testing.
I understand if you don't want to include that in the PR.
Then that would really need to be tackled as a separate PR.
You have a better grasp of the situation; as I said, that was just my impression.

bin/ncp/CONFIG/nc-datadir.sh
MagelanM: The same reasoning as in the previous comment—I felt it would be better to eliminate manual interventions entirely.

bin/ncp/CONFIG/nc-init.sh
MagelanM: In my recent tests, I noticed that the HPB service was always showing as down.
Once I restarted the service, everything worked fine.
I believe I identified two causes: first, the reload was missing in library.sh, and second, it wasn’t being called because the lock file (see condition) always existed at that point. The lock file couldn’t be removed earlier because other steps depend on it, so I postponed the initialization until the configuration was complete.

bin/ncp/NETWORKING/freeDNS.sh
MagelanM: That's right; this isn't directly related to the upgrade to Debian 13 either.
During testing, I ran into a problem where the dpkg process hadn't finished yet, but the next steps had already been triggered. I thought it would be a good idea to handle that centrally in library.sh.
However, with the FreeDNS installation script, it didn’t seem practical to use the function from library.sh, so this is actually redundant.

bin/ncp/NETWORKING/letsencrypt.sh
MagelanM: At this point, SSLUseStapling should be enabled because I disabled it globally in http2.conf.sh (it would have been better to do this in a commit)

etc/ncp-templates/mysql/91-ncp.cnf.sh
MagelanM: There is no direct connection to the upgrade to Debian 13.
I just happened to notice it while troubleshooting.
Currently, ncp is using MariaDB 11.8.6
innodb_large_prefix=true -> This parameter was removed starting with MariaDB 10.5. It is essentially always enabled, and I was advised to remove it just to be safe.
https://mariadb.com/docs/release-notes/community-server/old-releases/10.3/what-is-mariadb-103#system-variables

update.sh
MagelanM: During my testing of the PR, a hardcoding change was added for PHP 8.3. That was the only way I found to get it working again.
I thought it would also be good for the update, because it might need to be fixed in existing installations as well.

etc/ncp-templates/systemd/php-fpm.service.d.ncp.conf.sh
MagelanM: During my testing of the PR, a hardcoding change was added for PHP 8.3. That was the only way I found to get it working again.
I've tried a lot of different versions, and I'm still not sure.

bin/ncp/CONFIG/nc-datadir.sh
theCalcaholic: `occ` is not a daemon, it's basically active whenever we invoke it (and reads the config file whenever it needs it, so there's no "loading up" of any sorts).

However, some parts of ncp have to act in situations where PHP is not available, that's why we need fallbacks occasionally. I'm not sure if we need it here, but I would rather not find out via bug reports. :)

bin/ncp/NETWORKING/letsencrypt.sh
theCalcaholic: But that means, that it stays enabled when deactivating letsencrypt (that's what this case is, due to `"${ACTIVE}" != "yes"`). 🤔

theCalcaholic:
continued in #2123

bin/ncp/CONFIG/nc-datadir.sh
MagelanM: Yes.
The key point was to correct the order: config:system:set datadirectory before the mv command.
The sed fallbacks are a separate topic and do not need to be changed in this PR

bin/ncp/NETWORKING/letsencrypt.sh
MagelanM: You're right, that was a mistake in my PR and a misunderstanding of your question. It should be 'off' letsencrypt gets deactivate.

